PR #41, opened 2019-02-07
Fix for schema imports with new tests validating

Deeg-Kim:
*Description of changes:*

* Changed the parsing of id within import to be a string instead of symbol
* Added new test cases to verify that type imports are happening
 * Added new test cases to verify that imported types would be used during validation

### DIFF
--- a/data/test/schema/Customer.isl
+++ b/data/test/schema/Customer.isl
@@ -1,4 +1,7 @@
 schema_header::{
+    imports: [
+        { id: "schema/CommonTypes.isl", type: positive_int }
+    ],
 }
 
 type::{
@@ -8,6 +11,7 @@ type::{
     firstName: { type: string, occurs: required },
     middleName: string,
     lastName: { type: string, codepoint_length: range::[min, 7], occurs: required },
+    age: { type: positive_int, occurs: optional },
   }
 }
 
@@ -18,11 +22,13 @@ valid::{
   Customer: [
     { firstName: "Phil", lastName: "Collins" },
     { firstName: "Phil", lastName: "Collins", middleName: "Billy" },
+    { firstName: "Phil", lastName: "Collins", age: 68 },
   ],
 }
 
 invalid::{
   Customer: [
     { firstName: "Phil", middleName: "Billy" },
+    { firstName: "Phil", lastName: "Collins", age: -1 },
   ],
 }
--- a/test/software/amazon/ionschema/IonSchemaImportTest.kt
+++ b/test/software/amazon/ionschema/IonSchemaImportTest.kt
@@ -1,0 +1,43 @@
+package software.amazon.ionschema
+
+import org.junit.Assert.*
+import org.junit.Test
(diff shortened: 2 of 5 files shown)


test/software/amazon/ionschema/IonSchemaImportTest.kt
```diff
@@ -0,0 +1,43 @@
...
+                .loadSchema("schema/Customer.isl")
+                .getTypes()
+                .asSequence()
+                .associateBy({ it.name() })
+        assertTrue(types.contains("Customer"))
+        assertTrue(types.contains("positive_int"))
+    }
+
+    @Test
+    fun validateType() {
```
pbcornell: This appears to be redundant with the tests defined in data/test/schema/Customer.isl.  The preference here is for the file-based tests (so they can be leveraged to verify the behavior of other Ion Schema impls).  Are you seeing additional value in having this test method?

test/software/amazon/ionschema/IonSchemaImportTest.kt
```diff
@@ -0,0 +1,43 @@
...
+        val type = iss.loadSchema("schema/Customer.isl").getType("Customer")!!
+        assertTrue(type.isValid(ION.singleValue("{ firstName: \"Phil\", lastName: \"Collins\" }")))
+        assertFalse(type.isValid(ION.singleValue("{ firstName: \"Phil\", middleName: \"Billy\" }")))
+    }
+
+    @Test
+    fun validateAgainstImportedType() {
```
pbcornell: same comment as above for this test method

data/test/schema/Customer.isl
```diff
@@ -8,6 +11,7 @@ type::{
     firstName: { type: string, occurs: required },
     middleName: string,
     lastName: { type: string, codepoint_length: range::[min, 7], occurs: required },
+    age: { type: positive_int, occurs: optional },
```
pbcornell: `occurs: optional` is the default for a field.  Please change this line to `age: positive_int,`

test/software/amazon/ionschema/IonSchemaImportTest.kt
Deeg-Kim: Although we are effectively just testing the validation logic, I believe there is value in testing the validation against an imported type  -- this verifies that we do import, we can validate against it. More generally, this is a way to test that the type has been imported successfully.

I do agree that the test above may be less valuable.

pbcornell: Agreed, there's value in testing validation against an imported type.  However, the changes you've proposed in `data/test/schema/Customer.isl` do exactly that.  A successful test run against that file requires loading the Customer type, which in turn imports `positive_int` from `CommonTypes.isl`, and then verifying that the age field validates as expected.  It exercises the exact same set of API calls and codepaths that this test does.

Do we need unit tests that exercise API behavior?  Absolutely.  But we should avoid creating unit tests that duplicate those within `data/test`.

PR #79, opened 2019-03-09
adds API and tests against the API for review

pbcornell:

### DIFF
--- a/src/software/amazon/ionschema/IonSchemaSystem.kt
+++ b/src/software/amazon/ionschema/IonSchemaSystem.kt
@@ -1,0 +1,47 @@
+package software.amazon.ionschema
+
+import software.amazon.ion.IonValue
+
+/**
+ * Provides methods for instantiating instances of [Schema].
+ *
+ * To create an instance, use [IonSchemaSystemBuilder].
+ */
+interface IonSchemaSystem {
+    /**
+     * Requests each of the provided [Authority]s, in order, to resolve
+     * the requested schema id until one successfully resolves it.
+     *
+     * If an Authority throws an exception, resolution silently proceeds
+     * to the next Authority.
+     *
+     * @param[id] Identifier for the schema to load.
+     * @throws IonSchemaException if the schema id cannot be resolved.
+     * @throws InvalidSchemaException if the schema, once resolved, is determined to be invalid.
+     */
+    fun loadSchema(id: String): Schema
+
+    /**
+     * Constructs a new, empty schema.
+     *
+     * @return the new schema
+     */
+    fun newSchema(): Schema
+
+    /**
+     * Constructs a new schema using ISL provided as a String.
+     *
+     * @param[isl] ISL string representation of the schema to create
+     * @return the new schema
+     */
+    fun newSchema(isl: String): Schema
+
+    /**
+     * Constructs a new schema using ISL provided as Iterator<IonValue>.
+     *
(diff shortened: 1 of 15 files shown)


src/software/amazon/ionschema/Schema.kt
```diff
@@ -0,0 +1,45 @@
...
+     */
+    fun getSchemaSystem(): IonSchemaSystem
+
+    /**
+     * Constructs a new type using the type ISL provided as a String.
+     *
+     * @param[isl] ISL string representation of the type to create
+     * @return the new type
+     */
+    fun newType(isl: String): Type
```
tgregg: Same question regarding an `IonReader` variant that would interpret the value at which the given reader is positioned as a `Type`.

src/software/amazon/ionschema/Type.kt
```diff
@@ -0,0 +1,43 @@
...
+     * Returns the name of the type;  if the type has no name, a string representing
+     * the definition of the type is returned.
+     */
+    fun name(): String
+
+    /**
+     * If the specified value violates one or more of this type's constraints,
+     * returns `false`, otherwise `true`.
+     */
+    fun isValid(value: IonValue) = validate(this, value, true).isValid()
```
tgregg: I think it could also be useful to have `IonReader` variants of `isValid` and `validate` so that users that don't currently depend on the DOM don't need to start. Internally the implementation may still operate on DOM objects retrieved from the given reader, but it wouldn't have to.

src/software/amazon/ionschema/Violations.kt
```diff
@@ -0,0 +1,143 @@
...
+ * @property[index] Within a sequence, the index of the element this object corresponds to.
+ * @property[value] The child value of this object corresponds to.
+ */
+class ViolationChild internal constructor (
+        val fieldName: String? = null,
+        val index: Int? = null,
+        var value: IonValue? = null
+) : Violations() {
+
+    internal fun addValue(v: IonValue) {
```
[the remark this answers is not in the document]
pbcornell: In the common case, a ViolationChild represents a single nested value, the constraint(s) violated by the value, and any children of the child.

ViolationChild.addValue() exists solely for the less common case of an occurs constraint that allows multiple instances of a field.  If the occurs constraint is violated, the ViolationChild accumulates each instance of the field in its `value` property.  For an example of this in action, see the violations for children "a" and "c" in the [validation_fields](https://github.com/amzn/ion-schema-kotlin/blob/master/data/test/constraints/occurs/validation_fields.isl) test.

src/software/amazon/ionschema/IonSchemaSystem.kt
```diff
@@ -0,0 +1,47 @@
...
+     */
+    fun newSchema(): Schema
+
+    /**
+     * Constructs a new schema using ISL provided as a String.
+     *
+     * @param[isl] ISL string representation of the schema to create
+     * @return the new schema
+     */
+    fun newSchema(isl: String): Schema
```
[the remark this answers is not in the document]
pbcornell: Opened #90 to track this idea

src/software/amazon/ionschema/Schema.kt
pbcornell: Also covered by new issue #90

src/software/amazon/ionschema/Type.kt
pbcornell: Opened #91 to track this idea

PR #81, opened 2019-03-09
adds range impls and tests for review

pbcornell:

### DIFF
--- a/src/software/amazon/ionschema/internal/util/Range.kt
+++ b/src/software/amazon/ionschema/internal/util/Range.kt
@@ -20,7 +20,7 @@ internal enum class RangeType {
 /**
  * Interface for all range implementations.
  */
-internal interface Range<in T : Any> {
+internal interface Range<in T> {
     fun contains(value: T): Boolean
 }
 
@@ -29,7 +29,7 @@ internal interface Range<in T : Any> {
  */
 internal class RangeFactory {
     companion object {
-        fun <T : Any> rangeOf(ion: IonValue, rangeType: RangeType): Range<T> {
+        fun <T> rangeOf(ion: IonValue, rangeType: RangeType): Range<T> {
             if (ion.isNullValue) {
                 throw InvalidSchemaException("Invalid range $ion")
             }
--- a/src/software/amazon/ionschema/internal/util/Range.kt
+++ b/src/software/amazon/ionschema/internal/util/Range.kt
@@ -1,5 +1,6 @@
 package software.amazon.ionschema.internal.util
 
+import software.amazon.ion.IonInt
 import software.amazon.ion.IonList
 import software.amazon.ion.IonSymbol
 import software.amazon.ion.IonValue
@@ -28,25 +29,21 @@ internal interface Range<in T : Any> {
  */
 internal class RangeFactory {
     companion object {
-        @JvmStatic
         fun <T : Any> rangeOf(ion: IonValue, rangeType: RangeType): Range<T> {
             if (ion.isNullValue) {
                 throw InvalidSchemaException("Invalid range $ion")
             }
 
-            val ionList = if (ion !is IonList) {
+            val ionList = when (ion) {
+                !is IonList -> {
                     val range = ion.system.newList(ion.clone(), ion.clone())
                     range.addTypeAnnotation("range")
                     range
-                } else {
-                    ion
                 }
-
-            if (ionList.size != 2 || ionList[0].isNullValue || ionList[1].isNullValue
-                    || ((ionList[0] as? IonSymbol)?.stringValue() == "min"
-                            && ((ionList[1] as? IonSymbol)?.stringValue() == "max"))) {
-                throw InvalidSchemaException("Invalid range $ion")
+                else -> ion
             }
+
+            checkRange(ionList)
 
             @Suppress("UNCHECKED_CAST")
             return when (rangeType) {
@@ -59,4 +56,20 @@ internal class RangeFactory {
         }
     }
 }
+
+internal fun checkRange(ion: IonList) {
+    when {
+        !ion.hasTypeAnnotation("range") ->
+            throw InvalidSchemaException("Invalid range, missing 'range' annotation:  $ion")
+        ion.size != 2 ->
+            throw InvalidSchemaException("Invalid range, size of list must be 2:  $ion")
+        ion[0].isNullValue || ion[1].isNullValue || (isRangeMin(ion[0]) && isRangeMax(ion[1])) ->
+            throw InvalidSchemaException("Invalid range $ion")
+    }
+}
+
+internal fun isRangeMin(ion: IonValue) = (ion as? IonSymbol)?.stringValue().equals("min")
+internal fun isRangeMax(ion: IonValue) = (ion as? IonSymbol)?.stringValue().equals("max")
+
+internal fun toInt(ion: IonValue) = (ion as? IonInt)?.intValue()
 
--- a/src/software/amazon/ionschema/internal/util/RangeBigDecimal.kt
+++ b/src/software/amazon/ionschema/internal/util/RangeBigDecimal.kt
@@ -28,24 +28,16 @@ internal class RangeBigDecimal(private val ion: IonList) : Range<BigDecimal> {
     internal val lower: Boundary
     internal val upper: Boundary
     init {
-        if (!ion.hasTypeAnnotation("range")) {
-            throw InvalidSchemaException("Invalid range, missing 'range' annotation:  $ion")
-        }
-        if (ion.size != 2) {
-            throw InvalidSchemaException("Invalid range, size of list must be 2:  $ion")
-        }
+        checkRange(ion)
 
-        lower = if (ion[0] is IonSymbol && (ion[0] as IonSymbol).stringValue() == "min") {
-                Boundary(null, Infinity.NEGATIVE)
-            } else {
-                Boundary(ion[0], Infinity.NEGATIVE)
-            }
-        upper = if (ion[1] is IonSymbol && (ion[1] as IonSymbol).stringValue() == "max") {
-                Boundary(null, Infinity.POSITIVE)
-            } else {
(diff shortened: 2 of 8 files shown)


src/software/amazon/ionschema/internal/util/Range.kt
```diff
@@ -0,0 +1,62 @@
...
+                    val range = ion.system.newList(ion.clone(), ion.clone())
+                    range.addTypeAnnotation("range")
+                    range
+                } else {
+                    ion
+                }
+
+            if (ionList.size != 2 || ionList[0].isNullValue || ionList[1].isNullValue
+                    || ((ionList[0] as? IonSymbol)?.stringValue() == "min"
+                            && ((ionList[1] as? IonSymbol)?.stringValue() == "max"))) {
```
dlurton: Am I reading this correctly?  `range::[min, max]` is considered an invalid range?  It seems you also have asserting this to be invalid, however the specification has examples under the "Ranges" heading [here](https://amzn.github.io/ion-schema/docs/spec.html#constraints) that seem to suggest it should be valid.

dlurton: I looked some more and I found tests where `range::[min, max]` was considered valid and from a moderately quick read of the specification I don't see under what circumstances it would be valid or not...

src/software/amazon/ionschema/internal/util/RangeBigDecimal.kt
```diff
@@ -0,0 +1,120 @@
...
+                            throw InvalidSchemaException(
+                                    "Expected range lower/upper to be a decimal, float, or int (was $ion)")
+                    }
+                }
+    }
+
+    internal val lower: Boundary
+    internal val upper: Boundary
+    init {
+        if (!ion.hasTypeAnnotation("range")) {
```
dlurton: I'm not sure why none of the other `Range<T>` implementations do not have a similar check.  You can probably add this in the `init` block of a common base class which also defines a `protected val ion: IonList`) property.

src/software/amazon/ionschema/internal/util/RangeBigDecimal.kt
```diff
@@ -0,0 +1,120 @@
...
+            throw InvalidSchemaException("Invalid range, missing 'range' annotation:  $ion")
+        }
+        if (ion.size != 2) {
```
dlurton: same

src/software/amazon/ionschema/internal/util/RangeBigDecimal.kt
```diff
@@ -0,0 +1,120 @@
...
+            throw InvalidSchemaException("Invalid range, size of list must be 2:  $ion")
+        }
+
+        lower = if (ion[0] is IonSymbol && (ion[0] as IonSymbol).stringValue() == "min") {
```
dlurton: Slightly better:  

`(ion[0] as? IonSymbol)?.stringValue() == "min"`.  Comparing `null` to a string literal won't result in an NPE in Kotlin.

```
        val someList = listOf("foobar", null)
        someList.forEach {
            if(it == "foobar") println(it) else println("not foobar")
        }
```

src/software/amazon/ionschema/internal/util/RangeBigDecimal.kt
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    /**
+     * Represents either a lower or upper boundary of a range.
+     */
+    internal class Boundary(ion: IonValue?, private val infinity: Infinity) : Comparable<Boundary> {
+        internal val value: BigDecimal?
+        internal val boundaryType: RangeBoundaryType
+        init {
+            value = if (ion == null || ion.isNullValue) {
```
dlurton: I see this pattern being repeated a lot.  You can  add an extension property:

```
val IonValue?.isNullOrIonNull get() this == null || this.isNullValue
```
Or you can make it a function if you prefer:
```
fun IonValue?.isNullOrIonNull() = this == null || this.isNullValue

File: src/software/amazon/ionschema/internal/util/RangeInt.kt [hunk 6]
```diff
@@ -0,0 +1,44 @@
...
+ * Implementation of Range<Int>, which mostly delegates to RangeBigDecimal.
+ */
+internal class RangeInt (
+        private val ion: IonList,
+        private val delegate: RangeBigDecimal = RangeBigDecimal(ion)
+) : Range<Int> {
+
+    init {
+        if (!(ion[0] is IonInt
+                        || (ion[0] as? IonSymbol)?.stringValue().equals("min"))) {
```
dlurton: I see this being repeated enough I'd recommend an extension function:

```
fun IonValue?.isMinSymbol() = this as? IonSymbol?.stringValue() == "min"
```

Same for max.

src/software/amazon/ionschema/internal/util/RangeIntNonNegative.kt
```diff
@@ -0,0 +1,33 @@
...
+ * Implementation of Range<Int> restricted to non-negative integers.
+ * Mostly delegates to RangeInt.
+ */
+internal class RangeIntNonNegative (
+        private val ion: IonList,
+        private val delegate: RangeInt = RangeInt(ion)
+) : Range<Int> by delegate {
+
+    init {
+        if (!((ion[0] is IonInt && (ion[0] as IonInt).intValue() >= 0)
```
dlurton: This again feels like an opportunity to D.R.Y.

src/software/amazon/ionschema/internal/util/RangeIonTimestampPrecision.kt
```diff
@@ -0,0 +1,93 @@
...
+ * @see IonTimestampPrecision
+ */
+internal class RangeIonTimestampPrecision (
+        ion: IonList
+) : Range<IonTimestamp> {
+
+    private val delegate: Range<Int>
+
+    init {
+        if (!ion.hasTypeAnnotation("range")) {
```
dlurton: Can be moved to the `init` block of a base class containing the `ion` property so it can be shared among all `Range<>` implementations.

src/software/amazon/ionschema/internal/util/RangeIonTimestampPrecision.kt
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+        delegate = RangeFactory.rangeOf<Int>(intRangeIon, RangeType.INT)
+    }
+
+    override fun contains(value: IonTimestamp) = delegate.contains(IonTimestampPrecision.toInt(value))
+}
+
+internal enum class IonTimestampPrecision (val id: Int) {
+    year(-4),
```
dlurton: The reason for the negative id values is mysterious to me and should be removed or the reasons for them documented.

dlurton: These also do not follow the typical uppercase naming convention.

dlurton: Here's an alternative I extracted  from another project of ours and simplified:

```Kotlin
enum class DataType(val typeName: String) {
    BOOLEAN("boolean"), 
    SMALLINT("smallint"),
    INTEGER("integer"),
    FLOAT("float");
    
    companion object {
        private val LOOKUP =
            DataType.values()
                .map { it.typeName to it }
                .toMap()

        fun forTypeName(typeName: String): DataType? = LOOKUP[typeName]
    }
}
```

It's a little more work but prevents the need to violate coding guidelines and has the advantage that the textual representation is now decoupled from the enum value's name.

dlurton: You could also do this if you don't care about decoupling the names:

```Kotlin
    enum class DataType {
        BOOLEAN,
        SMALLINT,
        INTEGER,
        FLOAT;

        companion object {
            private val LOOKUP =
                DataType.values()
                    .map { it.name.toLowerCase() to it }
                    .toMap()

            fun forTypeName(typeName: String): DataType? = LOOKUP[typeName]
        }
    
```

src/software/amazon/ionschema/internal/util/Range.kt
```diff
@@ -0,0 +1,75 @@
...
+internal interface Range<in T : Any> {
+    fun contains(value: T): Boolean
+}
+
+/**
+ * Factory method for instantiating Range<T> objects.
+ */
+internal class RangeFactory {
+    companion object {
+        fun <T : Any> rangeOf(ion: IonValue, rangeType: RangeType): Range<T> {
```
dlurton: I think `: Any` in `<T: Any>` isn't doing anything.

src/software/amazon/ionschema/internal/util/Range.kt
```diff
@@ -0,0 +1,62 @@
...
+            }
+
+            @Suppress("UNCHECKED_CAST")
+            return when (rangeType) {
+                    RangeType.INT                     -> RangeInt(ionList)
+                    RangeType.INT_NON_NEGATIVE        -> RangeIntNonNegative(ionList)
+                    RangeType.ION_NUMBER              -> RangeIonNumber(ionList)
+                    RangeType.ION_TIMESTAMP           -> RangeIonTimestamp(ionList)
+                    RangeType.ION_TIMESTAMP_PRECISION -> RangeIonTimestampPrecision(ionList)
+                } as Range<T>
```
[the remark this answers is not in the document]
dlurton: This class is internal so it's not a huge deal but FWIW perhaps not using a factory at all would be a better approach.

src/software/amazon/ionschema/internal/util/RangeIonTimestampPrecision.kt
pbcornell: The coupling is intentional, and I don't see the value in decoupling in order to adhere to *guidelines*.

src/software/amazon/ionschema/internal/util/Range.kt
pbcornell: agreed, removed

PR #85, opened 2019-03-09
adds file-based tests, test runner, and README for review

pbcornell:


test/software/amazon/ionschema/IonSchemaTestRunner.kt
```diff
@@ -0,0 +1,221 @@
...
+            "data/test/constraints/not/empty_type.isl"
+    )
+
+    private val specialFieldNames = setOf("fields", "element")
+
+    override fun run(notifier: RunNotifier) {
+        val base = "data/test"
+        File(base).walk()
+            .filter { it.isFile }
+            .filter { !it.path.endsWith(".md") }
```
zslayton: Would it be better to filter out anything that doesn't have the `.isl` file extension? Currently that's just `README.md`, but you might have other files lying around in the future.

test/software/amazon/ionschema/IonSchemaTestRunner.kt
```diff
@@ -0,0 +1,221 @@
...
+                val testName = file.path.substring(base.length + 1, file.path.length - ".isl".length)
+                var schema: Schema? = null
+                var type: Type? = null
+
+                val iter = ION.iterate(FileReader(file)).asSequence().toList().listIterator()
+                iter.forEach { ion ->
+                    val annotation = ion.typeAnnotations[0]
+                    when (annotation) {
+                        "schema_header" -> {
+                            iter.previous()
```
zslayton: What's the benefit of defining type definition tests as a series of top-level values in a document? Since you're using the DOM everywhere already, a document like:
```
{
  "type" : {...},
  "valid" : [...],
  "invalid" : [...]
}
```
would allow you to parse the whole thing into an `IonStruct` and then request each value with `testDoc.get("type")`.

pbcornell: The struct approach you've illustrated would be ok for simple tests.  However, consider that the current representation:
* is valid ISL:  type definitions are top-level values, and other top-level values are open content
* more easily allows for definition of a schema (schema_header, _n_ type definitions, schema_footer) for testing purposes, instead of trying to represent a schema nested within a struct
* generally, my "test processor" mindset aligns more readily with "here's a sequence of steps to perform," instead of a group of key/value pairs;  at the top-most level, a document (or list) feels more appropriate (and yes, the same _could_ be accomplished via a struct;  for example, consider testing an HTTP endpoint:

```
{
  put: a,
  validate: a,
  put: b,
  validate: b,
  ...
}
```

you _could_, but i wouldn't recommend it)

I'll add that I'm pretty happy with the way this played out;  the ability for the test runner to simply switch on the annotation of the top-level value allowed adding new types of tests/processing quite easily.

pbcornell:
Change already applied in master; closing

PR #100, opened 2019-04-10
adds fun Schema.plusType(String | IonStruct): Schema

### LINKED ISSUE #73: add a type to a schema
Would "add a type to a schema" fall under #46 or #47?  I have a use case for building the schema dynamically; right now my workaround is to keep a list of type definitions myself and rebuild the `Schema` object every time (which is pretty slow -- O(N^2) for adding N types)

_Originally posted by @mtcalvin in https://github.com/amzn/ion-schema-kotlin/issues/47#issuecomment-470241040_

pbcornell:
Resolves #73 

Following the pattern of PCollection's `PMap.plus` and Kotlin's `Map.plus`, adds `Schema.plusType()` methods while maintaining Schema immutability.

### DIFF
--- a/src/software/amazon/ionschema/Schema.kt
+++ b/src/software/amazon/ionschema/Schema.kt
@@ -8,6 +8,14 @@ import software.amazon.ion.IonStruct
  * Each type may refer to other types within the same schema,
  * or types imported into this schema from other schemas.
  * To instantiate a Schema, see [IonSchemaSystem].
+ *
+ * Classes that implement this interface are expected to be
+ * immutable.  This avoids surprising behavior, for instance:
+ * if a particular type in a schema were allowed to be replaced,
+ * a value that was once valid for the type may no longer be valid.
+ * Instead, any methods that would mutate a Schema are expected
+ * to return a new Schema instance with the mutation applied
+ * (see [plusType] as an example of this).
  */
 interface Schema {
     /**
@@ -41,5 +49,23 @@ interface Schema {
      * @return the new type
      */
     fun newType(isl: IonStruct): Type
+
+    /**
+     * Constructs a new type using the type ISL provided as a String,
+     * and returns a new Schema instance containing all the types
+     * of this instance plus the new type.  Note that the new type
+     * in the returned instance will hide a type of the same name
+     * from this instance.
+     */
+    fun plusType(isl: String): Schema
+
+    /**
+     * Constructs a new type using the type ISL provided as an IonStruct.
+     * and returns a new Schema instance containing all the types
+     * of this instance plus the new type.  Note that the new type
+     * in the returned instance will hide a type of the same name
+     * from this instance.
+     */
+    fun plusType(isl: IonStruct): Schema
 }
 
--- a/src/software/amazon/ionschema/Schema.kt
+++ b/src/software/amazon/ionschema/Schema.kt
@@ -51,21 +51,11 @@ interface Schema {
     fun newType(isl: IonStruct): Type
 
     /**
-     * Constructs a new type using the type ISL provided as a String,
-     * and returns a new Schema instance containing all the types
-     * of this instance plus the new type.  Note that the new type
+     * Returns a new Schema instance containing all the types of this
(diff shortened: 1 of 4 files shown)


src/software/amazon/ionschema/Schema.kt
```diff
@@ -41,5 +49,23 @@ interface Schema {
...
+    fun plusType(isl: String): Schema
+
+    /**
+     * Constructs a new type using the type ISL provided as an IonStruct.
+     * and returns a new Schema instance containing all the types
+     * of this instance plus the new type.  Note that the new type
+     * in the returned instance will hide a type of the same name
+     * from this instance.
+     */
+    fun plusType(isl: IonStruct): Schema
```
zslayton: Accepting the raw components needed to construct a `Type` object (e.g. `String`, `IonStruct`) would seem to require duplicating the validation and processing logic for `String` and `IonStruct` in a few places, potentially increasing the number of code points where exceptions need to be handled.

Is it feasible to only offer a `fun plusType(t: Type): Schema` method and expect users to perform the preliminary conversion themselves, either in advance or inline (`plusType(Type(ionString)`)?

src/software/amazon/ionschema/internal/SchemaImpl.kt
```diff
@@ -5,25 +5,36 @@ import software.amazon.ion.IonString
...
  */
-internal class SchemaImpl(
+internal class SchemaImpl internal constructor(
         private val schemaSystem: IonSchemaSystemImpl,
         private val schemaCore: SchemaCore,
-        schemaContent: Iterator<IonValue>
+        schemaContent: Iterator<IonValue>,
+        /*
+         * [types] is declared as a MutableMap in order to be populated DURING
+         * INITIALIZATION ONLY.  This enables type B to find it's already-loaded
```
zslayton: nit: ~it's~ its

pbcornell: thank you!

src/software/amazon/ionschema/Schema.kt
pbcornell: Not sure what duplication you're referring to...?  The implementation of `plusType` calls `newType` to avoid duplication.

Yes, we could collapse the two `plusType` methods to just `fun plusType(t: Type): Schema` as you suggest.  I was expecting the current proposal to be simpler for callers, but it isn't, really;  consider:
```
var schema = ...
schema = schema.plusType("type::{ ... }")
val type = schema.getType(...)!!
```
vs.
```
var schema = ...
val type = schema.newType("type::{ ... }")
schema = schema.plusType(type)
```

It's nice to avoid the `!!`, but it seems odd to have to call `newType` and then `plusType`.  For those that like to chain, the first option could be written as:
```
var schema = ...
val type = schema.plusType("type::{ ... }").getType(...)!!
```

This isn't an option for the second style, but I don't think this offers a particularly strong argument either way.  Hmm...

zslayton: Ah, I didn't realize that a `Type` couldn't be constructed independent of a `Schema`. That does make things a bit verbose. I thought you could simply create a type with `Type(dataSource)`.

I was thinking of the case in which the `String` or `IonStruct` that's passed in is not a legal type definition. [The javadoc (kotlindoc?)](https://github.com/amzn/ion-schema-kotlin/blob/schema-add-type/src/software/amazon/ionschema/Schema.kt#L49) doesn't mention what will happen in the event of such a failure.

For example, I had expected the following to throw an exception that would need to be handled at the callsite:

```kotlin
try {
    val type = schema.newType("This clearly isn't an ISL type definition.")
} catch (... something ? ...) {
 //...
}
```

My suggestion was aiming to confine the `try/catch`es to the process of constructing a `Type` object, allowing them to be used freely thereafter. I see now that I misunderstood the design.

pbcornell: I see.  Ok, I've gone ahead and reduced this to a single `plusType(type: Type): Schema` method for the time being.

PR #101, opened 2019-04-10
fixes 'not: {}' test and removes it from the blacklist

### LINKED ISSUE #71: fix the "not: {}" test in data/test/constraints/not/not.isl
This test is currently commented out

pbcornell:
Resolves #71 

Test had been moved to data/test/constraints/not/empty_type.isl

Test required the default "type: any" behavior to be overridden in order to verify that "not: {}" (aka "not: { type: any }") is valid for null values.  That's why it was failing prior to this change.

### DIFF
--- a/data/test/constraints/not/empty_type.isl
+++ b/data/test/constraints/not/empty_type.isl
@@ -1,10 +1,34 @@
 type::{
   not: {},     // equivalent to not: { type: any }
+  type: $any,  // override the default "type: any" constraint, otherwise any null value will be invalid
 }
 valid::[
   null,
+  null.blob,
+  null.bool,
+  null.clob,
+  null.decimal,
+  null.float,
+  null.int,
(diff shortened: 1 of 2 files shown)


data/test/constraints/not/empty_type.isl
```diff
@@ -1,10 +1,34 @@
...
+  true,
   5,
+  5e0,
+  5d0,
+  2019-01-01T,
+  symbol,
+  "string",
+  {{ "clob" }},
+  {{aGVsbG8=}},
+  [],
```
wesboyt: it feels like empty values should return true?

pbcornell: an empty list (`[]`) is valid for `{ type: any }`, so it is invalid against `{ not: { type: any } }`.  does that help?

PR #119, opened 2019-06-05
implements the regex constraint with java.util.regex.Pattern while not exposing Pattern's extra features

### LINKED ISSUE #25: regex constraint:  avoid relying on Java's ScriptEngine
Java's built-in JavaScript engine provided a quick/dirty way to implement the regex constraint, but comes at a performance cost, and may provide features not expected by the Ion Schema Specification.  We should reimplement the regex constraint to align exactly with the spec, and not depend on Java's ScriptEngine.

pbcornell:
Resolves #25 


src/com/amazon/ionschema/internal/constraint/Regex.kt
```diff
@@ -21,55 +21,208 @@ import com.amazon.ion.IonValue
...
+                '+' -> error(si, "invalid character '$ch'")
+            }
+        }
+    }
+
+    private fun error(si: StringIterator, message: String): Unit =
+        throw InvalidSchemaException("$message in regex '$si' at offset ${si.index}")
+}
+
+private class StringIterator(private val s: String) {
```
therapon: probably more memory wasteful but did we consider `ArrayDequeue<Character>`?

src/com/amazon/ionschema/internal/constraint/Regex.kt
```diff
@@ -21,55 +21,208 @@ import com.amazon.ion.IonValue
...
+        if (sb.length > initialLength && ch != null) {
+            ch = si.peek()
+            when (ch) {
+                '?' -> error(si, "invalid character '$ch'")
+                '+' -> error(si, "invalid character '$ch'")
+            }
+        }
+    }
+
+    private fun error(si: StringIterator, message: String): Unit =
```
therapon: nit: I find it useful to add return types on method/function signatures. Once we make a decision then we should be consistent including methods that have no return value, i.e., `Unit`

src/com/amazon/ionschema/internal/constraint/Regex.kt
```diff
@@ -21,55 +21,208 @@ import com.amazon.ion.IonValue
...
+
+    fun next(): Char? {
+        index += 1
+        return get(index)
+    }
+
+    fun peek() = get(index + 1)
+
+    private fun get(i: Int): Char? {
+        if (i < length) {
```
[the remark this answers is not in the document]
pbcornell: `next()` always increments `index` before calling `get()`, and `peek()` always calls `get(index+1)`, so as written, `get()` should never receive a negative `i`

if other code decides to modify `index` directly, it's a different story; to protect against this, I've updated this class to make index private

src/com/amazon/ionschema/internal/constraint/Regex.kt
pbcornell: I've been declaring return types for public methods for clarity, but omitting them for anything internal.  In this instance, the compiler requires the return type.

src/com/amazon/ionschema/internal/constraint/Regex.kt
pbcornell: Hadn't considered it, but `ArrayDeque` is missing a way to determine the current index (well, i suppose queue length could be used for this).  Seems awkward and wasteful; sticking with `StringIterator`.

src/com/amazon/ionschema/internal/constraint/Regex.kt
```diff
@@ -21,55 +21,208 @@ import com.amazon.ion.IonValue
...
+            return s[i]
+        }
+        return null
```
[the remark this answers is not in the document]
pbcornell: I don't consider it an error to reach the end of a string (otherwise I agree an exception would make sense).  `null` seems to be a sufficient indicator of that condition for this use case.  And as noted previously, `get()` should always receive an `i` that is `>= 0`.

src/com/amazon/ionschema/internal/constraint/Regex.kt
therapon: Yes the compiler does require it. Which is why it makes me consider the option of adding it regardless. Internal developers (us) should get as much help as external developers. 

Not a deal breaker since we are already non-consistent.

src/com/amazon/ionschema/internal/constraint/Regex.kt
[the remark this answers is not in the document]
therapon: My concern was not with the end of the string but going beyond the end of the string, e.g., client code calls `next` a bunch of times and goes beyond the end of the string. 

With access to the fields being private this is not a big deal for this class in this context.

PR #142, opened 2020-05-14
fix(authorityfs): restricts file access to be within the specified basePath only

### LINKED ISSUE #132: Restrict AuthorityFilesystem to reading files within basePath
The current [AuthorityFilesystem](https://github.com/amzn/ion-schema-kotlin/blob/master/src/com/amazon/ionschema/AuthorityFilesystem.kt) implementation doesn't prevent callers from resolving a schemaId such as `../secret/xyz.isl`.  This is a security issue.  The implementation should be updated such that attempts to read files outside the basePath fail.

pbcornell:
Resolves #132 

* throws `FileNotFoundException` if requested to resolve a schema id outside the `basePath`
* also changed the exception to throw the more appropriate `FileNotFoundException` (instead of `IllegalArgumentException`)

### DIFF
--- a/src/com/amazon/ionschema/AuthorityFilesystem.kt
+++ b/src/com/amazon/ionschema/AuthorityFilesystem.kt
@@ -30,7 +30,7 @@ import java.io.Reader
  * @property[basePath] The base path in the filesystem in which to resolve schema identifiers.
  */
 class AuthorityFilesystem(basePath: String) : Authority {
-    private val basePath: String
+    private val canonicalBasePath: String
 
     init {
         val file = File(basePath)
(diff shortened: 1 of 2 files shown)


src/com/amazon/ionschema/AuthorityFilesystem.kt
```diff
@@ -28,15 +29,29 @@ import java.io.Reader
...
+        } else {
+            file.canonicalPath + File.separator
         }
     }
 
     override fun iteratorFor(iss: IonSchemaSystem, id: String): CloseableIterator<IonValue> {
         val file = File(basePath, id)
+
+        if (!file.canonicalPath.startsWith(basePath)) {
+            throw FileNotFoundException("$id (Permission denied)")
```
zslayton, line 52: Would [`AccessDeniedException`](https://docs.oracle.com/javase/7/docs/api/java/nio/file/AccessDeniedException.html) be a better fit for this?

There are some circumstances where `FileNotFoundException` doesn't strictly mean that the file wasn't found (e.g. [if you try to open a read-only file for writing](https://docs.oracle.com/javase/8/docs/api/index.html?java/io/FileNotFoundException.html)), but that would feel pretty surprising to me as a user.

pbcornell: Agreed, AccessDeniedException is a better fit; opting for [kotlin.io.AccessDeniedException](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/-access-denied-exception/) instead of the Java equivalent.  Thank you :)

PR #145, opened 2020-05-19
fix(validation): requires all top-level type definitions to have a name

### LINKED ISSUE #139: Require all top-level type definitions to have a name
As noted in #138, this library does not require top-level type definitions within a schema to have a name, and such types have an "implicit name" assigned by calling `toString()` on the type definition's in-memory representation (as an IonValue).  Allowing callers to retrieve an unnamed type by its "implicit name" is brittle at best, and is considered a mis-feature.

This library should:
1. require all top-level type definitions to have a name
1. not allow any type definitions to be retrieved by an "implicit name"

pbcornell:
*Description of changes:*
* requires all top-level type definitions to have a name (otherwise there's no good way to refer to them)
* I mistakenly pushed https://github.com/amzn/ion-schema-tests/commit/9937507ee8f11949e6f0489bee6f54b83c3f157f without a review; please comment on that as well
* the majority of test files in ion-schema-tests contain type definitions that don't have a name. these proposed changes include logic in ISLforISLTestRunner.kt to provide a type name, if absent; this allows all the ion-schema-tests test cases to pass without needing to provide a name for each type
* for backwards compatibility, if needed, provides an option to not throw on types that don't have a name

Resolves #139 

### DIFF
--- a/src/com/amazon/ionschema/IonSchemaSystemBuilder.kt
+++ b/src/com/amazon/ionschema/IonSchemaSystemBuilder.kt
@@ -39,6 +39,7 @@ class IonSchemaSystemBuilder private constructor() {
     private var constraintFactory = defaultConstraintFactory
     private var ionSystem = IonSystemBuilder.standard().build()
     private var schemaCache: SchemaCache? = null
+    private var params = mutableMapOf<IonSchemaSystemImpl.Param, Any>()
 
     /**
      * Adds the provided authority to the list of [Authority]s.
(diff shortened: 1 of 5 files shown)


src/com/amazon/ionschema/IonSchemaSystemBuilder.kt
```diff
@@ -81,6 +82,21 @@ class IonSchemaSystemBuilder private constructor() {
...
+     * Allows top-level types to not have a name.  Such types can't be referred to
+     * by name and are thereby of limited (if any?) value.  This option if provided
+     * in case consumers have defined top-level types that don't have names.
+     * Should only be used if needed for backwards compatibility with v1.0; this will
+     * be removed in a future release.
+     *
+     * @since 1.1
+     */
+    @Deprecated("For backwards compatibility with v1.0")
+    fun allowUnnamedTopLevelTypes(): IonSchemaSystemBuilder {
```
therapon, line 95: nit: unnamed = anonymous, rename to `allowAnonymousTopLevelTypes()`

pbcornell: agreed, adds clarity, changed, thank you

Issue #3, opened 2018-10-26
add support for open/closed content


pbcornell:
Per the specification, enable support for open content where needed, and enforce closed content when a type specifies it.

PR #148, opened 2020-05-22
feat(api): adds API for retrieving the types imported by a schema

### LINKED ISSUE #137: Provide an API for retrieving the imports of a Schema
Note that the API should allow for imports of a complete schema, a single type, multiple imports that import different types from the same schemaId, and type aliasing.

pbcornell:
Resolves #137 

#143 enabled callers to retrieve the ISL underlying Schema and Type objects.  This PR enables callers to retrieve schemas and types *imported* by a schema (and thereby gain access to the underlying ISL of those objects).

Prior to merging:  update the ion-schema-tests submodule after merging https://github.com/amzn/ion-schema-tests/pull/3.

### DIFF
--- a/src/com/amazon/ionschema/Import.kt
+++ b/src/com/amazon/ionschema/Import.kt
@@ -1,0 +1,45 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.ionschema
+
+/**
+ * An Import represents all the types imported by a Schema
+ * from one schema id.
+ *
+ * Note that multiple ISL imports referencing the same schema id
+ * (each importing/aliasing an individual type) are represented by a
+ * single Import instance.
+ */
+interface Import {
+    /**
+     * The schema id referenced by the import.
+     */
+    val id: String
+
+    /**
+     * Returns the requested type, if present in this import;
+     * otherwise returns null.  If a type is aliased (via "as")
+     * by the import, the type will only be returned from this
+     * method by its alias (not the imported type's original name).
+     */
+    fun getType(name: String): Type?
+
+    /**
+     * Returns an iterator over the types imported by this Import.
+     */
+    fun getTypes(): Iterator<Type>
+}
+
--- a/src/com/amazon/ionschema/Import.kt
+++ b/src/com/amazon/ionschema/Import.kt
@@ -25,9 +25,9 @@ package com.amazon.ionschema
  */
 interface Import {
     /**
-     * The schema id referenced by the import.
+     * Returns the schema referenced by the import.
      */
-    val id: String
+    fun getSchema(): Schema
 
     /**
      * Returns the requested type, if present in this import;
@@ -39,6 +39,8 @@ interface Import {
 
     /**
      * Returns an iterator over the types imported by this Import.
+     * Callers must not rely on any particular ordering, as Types
+     * may be returned in any order.
      */
     fun getTypes(): Iterator<Type>
 }
--- a/src/com/amazon/ionschema/Import.kt
+++ b/src/com/amazon/ionschema/Import.kt
@@ -25,6 +25,11 @@ package com.amazon.ionschema
  */
 interface Import {
     /**
+     * The id of the referenced schema.
+     */
+    val id: String
+
+    /**
      * Returns the schema referenced by the import.
      */
     fun getSchema(): Schema
--- a/src/com/amazon/ionschema/internal/ImportImpl.kt
+++ b/src/com/amazon/ionschema/internal/ImportImpl.kt
@@ -1,3 +1,18 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
 package com.amazon.ionschema.internal
 
 import com.amazon.ionschema.Import
--- a/src/com/amazon/ionschema/internal/ImportImpl.kt
+++ b/src/com/amazon/ionschema/internal/ImportImpl.kt
@@ -1,6 +1,7 @@
 package com.amazon.ionschema.internal
 
 import com.amazon.ionschema.Import
+import com.amazon.ionschema.Schema
 import com.amazon.ionschema.Type
 
 /**
@@ -8,11 +9,14 @@ import com.amazon.ionschema.Type
  */
 internal class ImportImpl(
         override val id: String,
-        private val types: Map<String, Type>
+        private val schema: Schema?,
+        private val types: Map<String, Type> = emptyMap()
 ) : Import {
 
-    override fun getType(name: String) = types[name]
+    override fun getType(name: String) = schema?.getType(name) ?: types[name]
 
-    override fun getTypes() = types.values.iterator()
+    override fun getTypes() =
+            ((schema?.getTypes()?.asSequence() ?: emptySequence())
+                    + types.values.asSequence()).iterator()
 }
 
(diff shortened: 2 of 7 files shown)


src/com/amazon/ionschema/Import.kt
```diff
@@ -0,0 +1,52 @@
...
+     */
+    fun getSchema(): Schema
+
+    /**
+     * Returns the requested type, if present in this import;
+     * otherwise returns null.  If a type is aliased (via "as")
+     * by the import, the type will only be returned from this
+     * method by its alias (not the imported type's original name).
+     */
+    fun getType(name: String): Type?
```
therapon, line 43: Given that we have `getType(name:String)` it will be straight forward to implement a `containsType(name:String): bool` predicate. Does it make it easier to provide that method as well on this interface?

src/com/amazon/ionschema/Import.kt
```diff
@@ -0,0 +1,52 @@
...
+     * Returns the schema referenced by the import.
+     */
+    fun getSchema(): Schema
+
+    /**
+     * Returns the requested type, if present in this import;
+     * otherwise returns null.  If a type is aliased (via "as")
+     * by the import, the type will only be returned from this
+     * method by its alias (not the imported type's original name).
+     */
```
therapon, line 42: For the argument `name` which name do I use the original type's name or the alias name or can I use either? Adding a comment here to make it clear would be good.

src/com/amazon/ionschema/Import.kt
```diff
@@ -0,0 +1,52 @@
...
+    fun getType(name: String): Type?
+
+    /**
+     * Returns an iterator over the types imported by this Import.
+     * Callers must not rely on any particular ordering, as Types
```
therapon, line 47: Would it be better to return a Set instead. Also the Types returned they hold the alias names from the import (`as: x`). 

If `Type` allows me to extract both names (original and aliased) then please ignore my question. 

I am mostly asking to make sure we document what these names are so as to not surprise client code.

src/com/amazon/ionschema/internal/ImportImpl.kt
```diff
@@ -0,0 +1,55 @@
...
+import com.amazon.ionschema.Type
+import java.lang.IllegalArgumentException
+
+/**
+ * Implementation of [Import] for all user-provided ISL.
+ */
+internal class ImportImpl(
+        override val id: String,
+        private val schema: Schema,
+        private val importEntireSchema: Boolean,
```
therapon, line 29: Can we have `importEntireSchema = true` and `types` not empty and not `null`? Which .isl file as input would cause such a scenario?

src/com/amazon/ionschema/Import.kt
pbcornell: > If a type is aliased (via "as") by the import, the type will only be returned from this method by its alias (not the imported type's original name).

I think this text is sufficient; how would you clarify further?

src/com/amazon/ionschema/Import.kt
pbcornell: My sense is that `containsType(key)` is not a great pattern.  I suspect typical usage of that method would be followed by `getType(key)` -- at which point, you've performed two look-ups instead of just one (plus a null check).

src/com/amazon/ionschema/Import.kt
pbcornell: I'm following the example set by [Schema.getTypes()](https://github.com/amzn/ion-schema-kotlin/blob/master/src/com/amazon/ionschema/Schema.kt#L51), as well as the new `Schema.getImports(): Iterator<Import>` method added in this PR.

For Set-like behavior, I'd expect an `Import` object's `getType(name)` method to be sufficient (the Import object effectively *is* the Set).  Whereas the style of `getTypes()` assumes you want to iterate over the types.

As for aliasing, the approach is to only expose an aliased type by its alias, not its original name.  So:

```
val type = import.getType(name)
assertEquals(name, type.name)
```

There is a catch, however.  The `IonValue` provided by `type.isl` *does not* take aliasing into account--it is the original ISL of the type.

src/com/amazon/ionschema/internal/ImportImpl.kt
pbcornell: Hopefully not common usage, but [schema/import/import_types.isl](https://github.com/amzn/ion-schema-tests/blob/master/schema/import/import_types.isl) provides an example of this.  The entire positive_int.isl schema is imported, *and* the `positive_int` type is aliased to `posint` via a second import.  These two imports result in invoking this constructor with `importEntireSchema = true` and a non-empty `types` map.

src/com/amazon/ionschema/Import.kt
therapon: This is not a blocker and I believe touches on personal preference on APIs. 

Well, if there is a  `containsType` method then the next step is to make the return type of `getType` be `Type` rather than `Type?`. If people call `getType` without first using `containsType` then you can throw.  We still have 2 lookups which may be a cost, but, we no longer have `null` as a possible return value and a check for `null` which may or may not happen on the client side probably more than once (if I chain calls or pass the return value around). 

I may be a little too eager in removing `null` from the API. As I said, not blocking, mostly exploring the API design space.

src/com/amazon/ionschema/Import.kt
therapon: I miss-read the comment. Sorry

src/com/amazon/ionschema/internal/ImportImpl.kt
therapon: Hm, so if for the `importEntireSchema` case we implemeted it as if we added 'as' statements to alias to the same name then the implementation will only have to deal with aliases and should simplify the code. Unless there is a need to be aware that an import did not add aliases at all for some other reason

therapon:
> There are two design considerations in this PR I should have hlghlighted earlier:
> 
>     1. Behavior of the existing `Schema.getType()` and `Schema.getTypes()` provides a unified/flattened (post-import?) view of a schema, whereas `Schema.isl`, `Schema.getImports()`, and `Schema.getImport()` expose the original, decomposed underpinnings of a Schema.  How concerning is this?  Is there a better way to model this in an API?
> 

If I understand correctly, I think this is unavoidable, right? By explicitly asking to get and inspect the imports of a schema, I infer that I do want to inspect the decomposed underpinnings of this Schema. I was mostly reviewing the code to see that the APIs provided the underpinnings in a systematic way. 


>     2. An Import object corresponding to a partial schema import provides access to only the imported/aliased types via `Import.getType()` and `Import.getTypes()`, but it also provides access to the _entire_ Schema via `Import.getSchema()`.  This enables use cases which require access to open content in the schema_header of an imported type's schema, but also provides easy access to _all_ of the types, even if only one was imported (via `Import.getSchema().getType("non-imported type")`).  How concerning is this?

I am assuming that there is a use case (open content) that makes this a feature, so I have no concerns with the feature being added. My comments were from the point of view of API provided for the feature (is it only aliased names or both etc.)

src/com/amazon/ionschema/Import.kt
pbcornell: Yeah, feels too eager to me.

The nice thing is that checking for null after calling `getType(name): Type?` is optional.  If you don't check in Java, you may wind up with an NPE.  In Kotlin, simply append a not-null assertion (`!!`).

Leaving as-is.  We can narrow the return type (to exclude `null`) later if desired, but there's something to be said for following the pattern of platform APIs like `Map.get(key: K): V?`.

pbcornell:
> > There are two design considerations in this PR I should have hlghlighted earlier:
> > ```
> > 1. Behavior of the existing `Schema.getType()` and `Schema.getTypes()` provides a unified/flattened (post-import?) view of a schema, whereas `Schema.isl`, `Schema.getImports()`, and `Schema.getImport()` expose the original, decomposed underpinnings of a Schema.  How concerning is this?  Is there a better way to model this in an API?
> > ```
> 
> If I understand correctly, I think this is unavoidable, right? By explicitly asking to get and inspect the imports of a schema, I infer that I do want to inspect the decomposed underpinnings of this Schema. I was mostly reviewing the code to see that the APIs provided the underpinnings in a systematic way.
> 

I'm reminded of an API from a previous life that provided composed and decomposed views of underlying data.  I think it was warranted in that case, but certainly added complexity.  Not sure whether such an approach will make sense for this API at some point, but your comments suggest to me that we're not there yet, and that the behavior as described/implemented here is not unexpected.

> > ```
> > 2. An Import object corresponding to a partial schema import provides access to only the imported/aliased types via `Import.getType()` and `Import.getTypes()`, but it also provides access to the _entire_ Schema via `Import.getSchema()`.  This enables use cases which require access to open content in the schema_header of an imported type's schema, but also provides easy access to _all_ of the types, even if only one was imported (via `Import.getSchema().getType("non-imported type")`).  How concerning is this?
> > ```
> 
> I am assuming that there is a use case (open content) that makes this a feature, so I have no concerns with the feature being added. My comments were from the point of view of API provided for the feature (is it only aliased names or both etc.)

Yes, feature as it provides access to the ISL (and thereby open content) of an imported schema/type.  Alternatively, `Import.getSchema()` could return a view of the Schema that only exposes the imported types (and `Import.getSchema().getIsl()` would construct an IonValue that only represents the imported types...?).  Or instead of `Import.getSchema()`, we provide `Import.getSchemaIsl(): IonDatagram`....

src/com/amazon/ionschema/internal/ImportImpl.kt
pbcornell: I like the idea of fetching types from the Schema (as currently implemented in this PR); if at some point we allow types to be added to a Schema, a full import of the Schema would reflect those changes.

However, your comment has made me realize this approach presents a potential problem.  If my schema defines type A and imports all of schema X, if someone can dynamically add a new type A to schema X, which type A does my schema see?  We've avoided such problems up to this point, and the approach you're describing avoids inadvertently presenting this issue in the future.  Will take a closer look.

pbcornell: Ok, I've simplified as you suggested, and `importEntireSchema` is no more.  This feels all-around safer and a bit simpler.

Issue #58, opened 2019-02-28
stack overflow occurs when two schemas import each other


pbcornell:

popematt:
Fixed by #157

PR #157, opened 2020-10-14
cyclic dependencies issue

desaikd:
*Issue #58*

*Description of changes:*
This is a draft PR for #58.

**Tasks:**

1. Added schemaImportSet in IonSchemaSystemImpl to store all imports in order to detect cycle.
2. In schemaImportSet added 2 new variables nocycle(stores whether cycle exists or not) and normalizedId(stores a normalized schema Id with path of the schema to be imported)
3. Throws error if schema tries to import itself.
4. loadHeader if cycle is detected will not load schema again to break the dependency/import cycle.

**Questions:**

1. Is there another way to find id for Schema other than schemaId? In order to always get the relative path to schema file(`schemaId`) rather then a [testName](https://github.com/amzn/ion-schema-kotlin/blob/master/test/com/amazon/ionschema/IonSchemaTestRunner.kt#L63).
2. following up with above question if `normalizedId` looks like a good idea to store schema Ids for `importSet`?
3. Also, Should I add some test files for this issue?

### DIFF
--- a/src/com/amazon/ionschema/internal/SchemaImpl.kt
+++ b/src/com/amazon/ionschema/internal/SchemaImpl.kt
@@ -60,8 +60,6 @@ internal class SchemaImpl private constructor(
 
     private val imports: Map<String, Import>
 
-    var nocycle = false // this flag helps in not loading types when there is a cycle
-
     init {
         val dgIsl = schemaSystem.getIonSystem().newDatagram()
 
@@ -96,9 +94,8 @@ internal class SchemaImpl private constructor(
             if (!foundHeader && foundFooter) {
                 throw InvalidSchemaException("Found a schema_footer, but not a schema_header")
             }
-            if(nocycle) {
-                resolveDeferredTypeReferences()
-            }
+
+            resolveDeferredTypeReferences()
             imports = importsMap
 
         } else {
@@ -140,20 +137,12 @@ internal class SchemaImpl private constructor(
                 val alias = it["as"] as? IonSymbol
                 // if importSet has an import with this id then do not load schema again to break the cycle.
                 if(!importSet.contains(id.stringValue())) {
-                    nocycle = true
-
-                    // normalize schemaId to have .isl at the end as importSet need a common pattern to be followed
-                    // for Ids added into the set
-                    var normalizedId = ""
-                    if(schemaId != null) {
-                        normalizedId = if(schemaId.endsWith(".isl")) schemaId else schemaId + ".isl"
-                    }
+                    var normalizedId = schemaId?: "";
 
                     // if Schema is importing itself then throw error
                     if(normalizedId.equals(id.stringValue())) {
                         throw InvalidSchemaException("Schema can not import itself.")
                     }
-
 
                     // add parent and current schema to importSet and continue loading current schema
                     schemaSystem.addToSchemaImportSet(normalizedId)
--- a/src/com/amazon/ionschema/internal/SchemaImpl.kt
+++ b/src/com/amazon/ionschema/internal/SchemaImpl.kt
@@ -20,6 +20,7 @@ import com.amazon.ion.IonList
 import com.amazon.ion.IonString
 import com.amazon.ion.IonStruct
 import com.amazon.ion.IonSymbol
+import com.amazon.ion.IonText
 import com.amazon.ion.IonValue
 import com.amazon.ionschema.Import
 import com.amazon.ionschema.InvalidSchemaException
@@ -135,7 +136,7 @@ internal class SchemaImpl private constructor(
         (header.get("imports") as? IonList)
             ?.filterIsInstance<IonStruct>()
             ?.forEach {
-                val id = it["id"] as IonString
+                val id = it["id"] as IonText
                 val alias = it["as"] as? IonSymbol
                 // if importSet has an import with this id then do not load schema again to break the cycle.
                 if(!importSet.contains(id.stringValue())) {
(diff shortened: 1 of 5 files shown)


zslayton:
Reviewing [the spec](https://amzn.github.io/ion-schema/docs/spec.html#schema-authorities), it looks like the meaning of an import `id` is decided entirely by the schema authority responsible for performing the import.

> The structure of a id string (per the example above) is defined by the schema authority responsible for the schema/type(s) being imported. Note that runtime resolution of a schema over a network presents availability and security risks, and should thereby be avoided.
> 
> When resolving a schema, authorities may choose to follow well-known patterns; for example:
> 
> * a filesystem authority might specify that an id string corresponds to an ISL file relative to some base, e.g.: "{base}/com/example/core/Customer.isl"
> * a REST authority might specify that a id string is a resource URL that corresponds to an ISL file, e.g.: "https://{host}:{port}/{base}/com/example/core/Customer" (again, note the inherent availabilty and security risks here)

Filesystems have aliases (e.g. symlinks) and multiple URLs can easily point to the same resource, so unfortunately import `id`s seem to be a bit of a misnomer.

Off-hand, it seems we would need to either:

* Amend the spec to require that schema authorities forbid aliasing

OR

* Require that schema authority implementations provide a mechanism to generate a true identifier. (e.g. A filesystem schema authority could map the `id`s `/path/to/foo.isl` and `/path/to/foo_alias.isl` to the same true identifier, making it possible to dedup them.)

I think my preference is for the former, but I think it may be too late for that; from what I understand the filesystem authority already allows you to specify either `myfile` or `myfile.isl`.

src/com/amazon/ionschema/internal/SchemaImpl.kt
```diff
@@ -129,35 +128,51 @@ internal class SchemaImpl private constructor(
...
             ?.forEach {
                 val id = it["id"] as IonText
-                val importedSchema = schemaSystem.loadSchema(id.stringValue())
-                val schemaAndTypes = importsMap.getOrPut(id.stringValue()) {
-                    SchemaAndTypeImports(id.stringValue(), importedSchema)
-                }
+                val alias = it["as"] as? IonSymbol
+                // if importSet has an import with this id then do not load schema again to break the cycle.
+                if(!importSet.contains(id.stringValue())) {
+                    var normalizedId = schemaId?: "";
```
zslayton, line 140: Since we're no longer allowing/handling aliases, do we still need a `normalizedId`? Maybe instead of `normalizedId` and `id`  we could use names like `parentImportId` and `childImportId`?

src/com/amazon/ionschema/internal/SchemaImpl.kt
```diff
@@ -129,35 +128,51 @@ internal class SchemaImpl private constructor(
...
-                val typeName = (it["type"] as? IonSymbol)?.stringValue()
-                if (typeName != null) {
-                    var newType = importedSchema.getType(typeName)
-                            ?: throw InvalidSchemaException(
-                                "Schema $id doesn't contain a type named '$typeName'")
+                    // add parent and current schema to importSet and continue loading current schema
+                    schemaSystem.addToSchemaImportSet(normalizedId)
+                    schemaSystem.addToSchemaImportSet(id.stringValue())
+                    val importedSchema = schemaSystem.loadSchema(id.stringValue())
+                    schemaSystem.resetSchemaImportSet(id.stringValue(), normalizedId)
```
zslayton, line 151: There's a small incongruity here: we add the parent ID and child ID as two separate calls:

```kotlin
schemaSystem.addToSchemaImportSet(normalizedId)
schemaSystem.addToSchemaImportSet(id.stringValue())
```

but then remove them in a single call:

```kotlin
schemaSystem.resetSchemaImportSet(id.stringValue(), normalizedId)
```

Do we need the helper methods at all? Since it's an internal-only data structure, could we just use: 
```kotlin
importSet.add(parentId);
importSet.add(childId);
// ... do things
importSet.remove(parentId);
importSet.remove(childId);
```
?

desaikd: Do you mean setting importset to be `public` and directly using its `add` and `remove` methods?

zslayton: I don't think you need to make it `public`, do you? You [already have a handle on it](https://github.com/amzn/ion-schema-kotlin/blob/a5119e30891c291a09b4645c00b00c21bbb33b80/src/com/amazon/ionschema/internal/SchemaImpl.kt#L131):

```kotlin
val importSet: MutableSet<String> = schemaSystem.getSchemaImportSet()
```

src/com/amazon/ionschema/internal/SchemaImpl.kt
```diff
@@ -129,35 +126,52 @@ internal class SchemaImpl private constructor(
...
+                val childImportId = it["id"] as IonText
+                val alias = it["as"] as? IonSymbol
+                // if importSet has an import with this id then do not load schema again to break the cycle.
+                if(!importSet.contains(childImportId.stringValue())) {
+                    var parentImportId = schemaId?: "";
+
+                    // if Schema is importing itself then throw error
+                    if(parentImportId.equals(childImportId.stringValue())) {
+                        throw InvalidSchemaException("Schema can not import itself.")
+                    }
```
zslayton, line 143: Interesting corner case here: given the intentionally open-ended language in the spec, I think it's possible for the empty string to be a valid import `id`. `parentImportId` can be set to `""` on line 138; if that happens, attempting to import a child with `id=""` would cause an error.

To understand this better we'd need to look at all the invocation sites of this class's constructor to see where/why `schemaId` is set to `null`.

This doesn't need to hold up the PR, I'm adding this comment so I can easily open an issue for this. :)

zslayton (approved):

src/com/amazon/ionschema/internal/SchemaImpl.kt
therapon: The grammar does indeed have `ID` as either a `STRING` or `SYMBOL` both of which allow for the empty value. That is unfortunate in my opinion :)

therapon (approved):

PR #159, opened 2020-10-29
Implement 'closed' annotations (amzn/ion-schema#26)

mwfpope:
*Issue #, if available:*

Implements the spec change given in amzn/ion-schema#26

*Description of changes:*

Implements 'closed' annotations. When a schema type has an `annotations` constraint and the list of type annotations is annotated with `closed::`, IonValues will be validated as invalid if they are annotated with anything _not_ in the given list of annotations.

### DIFF
--- a/src/com/amazon/ionschema/internal/constraint/Annotations.kt
+++ b/src/com/amazon/ionschema/internal/constraint/Annotations.kt
@@ -72,7 +72,9 @@ internal class OrderedAnnotations(
     private val stateMachine: StateMachine
 
     init {
-        val stateMachineBuilder = StateMachineBuilder().withOpenContent()
+        val stateMachineBuilder = StateMachineBuilder().apply {
+            if (!ion.hasTypeAnnotation("closed")) withOpenContent()
+        }
         var state: State? = null
         (ion as IonList).forEachIndexed { idx, it ->
             val newState = State(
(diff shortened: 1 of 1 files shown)


src/com/amazon/ionschema/internal/constraint/Annotations.kt
```diff
@@ -106,6 +108,8 @@ internal class UnorderedAnnotations(
         private val annotations: List<Annotation>
 ) : ConstraintBase(ion) {
 
+    private val closedAnnotationStrings: List<String>? = if (ion.hasTypeAnnotation("closed")) (ion as IonList).map { (it as IonSymbol).stringValue() } else null
```
therapon, line 111: nit: Could we use an empty list instead of `null` here.

mwfpope: There's a few reasons why I think it's okay to use a nullable type here.

1. `null` is not an anti-pattern in Kotlin because it's part of the type system. I know this package is also consumed by clients using Java, but this particular field is a `private val` that clients of the library will never see.
2. In this `val`, `null` and the empty list mean two different things. If `closedAnnotationStrings` is `null`, it means that the annotation constraint is not `closed::`. If `closedAnnotationStrings ` is an empty list, it means that the annotation contraint is `closed::` and the list of allowed annotations is empty. (Cf. L126 on the "new" side of the diff.)

Do you think that is reasonable justification for using a nullable type here? I can add a comment to clarify why `null` and the empty list are distinct here if you think that would be helpful.

However, if you still prefer not using a nullable type, let me know and I will come up with a different way of representing this information.

therapon: I did not refer to an antipattern and the fact that `null` is in the language is not a reason for it to be a good pattern :) 

My nit was mainly because we are adding more context to what `null` means here. The cost of an extra private field, a boolean since it seems like we have 2 states, seems pretty minimal to me as a cost and gives me the benefit or clear code. 

As I said though this is a nit not a blocker for me.

therapon (approved):

PR #160, opened 2020-10-29
Updated testSuite to pass filename as id instead of testName

desaikd:
*Description of changes:*
In order to solve the cyclic dependency issue (specified by [this PR](https://github.com/amzn/ion-schema-kotlin/pull/157)), we have to change the way the schemaId is passed in the testSuite.
 
**Changes:**
1. Instead of using the `testName` stating file path without  '.isl', will be using `testName` with the '.isl' inside `IonSchemaTestRunner`. 
2. This will allow the cyclic dependency check to have a unique Id pattern followed, which can help in defining equality of import ids.

### DIFF
--- a/test/com/amazon/ionschema/IonSchemaTestRunner.kt
+++ b/test/com/amazon/ionschema/IonSchemaTestRunner.kt
@@ -60,7 +60,8 @@ class IonSchemaTestRunner(
             .filter { it.path.endsWith(".isl") }
             .filter { !blacklist.contains(it.path) }
             .forEach { file ->
-                val testName = file.path.substring(base.length + 1, file.path.length)
+                val testName = file.path.substring(base.length + 1, file.path.length - ".isl".length)
+                val testFile = file.path.substring(base.length + 1, file.path.length)
                 var schema: Schema? = null
                 var type: Type? = null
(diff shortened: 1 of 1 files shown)


test/com/amazon/ionschema/IonSchemaTestRunner.kt
```diff
@@ -60,7 +60,7 @@ class IonSchemaTestRunner(
             .filter { it.path.endsWith(".isl") }
             .filter { !blacklist.contains(it.path) }
             .forEach { file ->
-                val testName = file.path.substring(base.length + 1, file.path.length - ".isl".length)
+                val testName = file.path.substring(base.length + 1, file.path.length)
```
zslayton, line 63: I think this will cause all of the test names in the output to include the `.isl` suffix as well. If you make a separate value (`testFile`?) and pass that to the `SchemaImpl` constructors downstream, we can update the tests' behavior without changing their names.

```suggestion
                val testFile = file.path.substring(base.length + 1, file.path.length)
                val testName = file.path.substring(base.length + 1, file.path.length - ".isl".length)
```

zslayton (approved):

Issue #9, opened 2018-10-26
assert that constraints are compatible with the type on which they are specified


pbcornell:
Consider the following type definition, for which there is no valid value:  `type::{ type: int, codepoint_length: 5 }`.  It would be helpful to indicate that this is an invalid type definition (or at least provide a warning) when a constraint is specified on a type definition it is not compatible with (in this example, a value of type 'int' doesn't have any codepoints, so codepoint_length will always be false).

PR #164, opened 2021-02-17
Add changes for `float_size` constraint

desaikd:
*Issue https://github.com/amzn/ion-schema/issues/18*

*Description of changes:*
This PR adds changes for `float_size` constraint

*Test*:
https://github.com/amzn/ion-schema-tests/pull/9

### DIFF
--- a/src/com/amazon/ionschema/internal/constraint/FloatSize.kt
+++ b/src/com/amazon/ionschema/internal/constraint/FloatSize.kt
@@ -1,0 +1,59 @@
+package com.amazon.ionschema.internal.constraint
+
+
+import com.amazon.ion.IonFloat
+import com.amazon.ion.IonSymbol
+import com.amazon.ion.IonValue
+import com.amazon.ionschema.InvalidSchemaException
+import com.amazon.ionschema.Violation
+import com.amazon.ionschema.Violations
+import java.math.BigDecimal
+
+/**
+ * Implements the float_size constraint.
+ */
+internal class FloatSize(
+        ion: IonValue
+) : ConstraintBase(ion) {
+
+    private val validValues = setOf("float16", "float32", "float64")
+    private val maxFloat32Range: BigDecimal = BigDecimal.valueOf(Float.MAX_VALUE.toDouble())
+    private val minPosValue: BigDecimal = BigDecimal.valueOf(Float.MIN_VALUE.toDouble())
+
+
+    init {
+        if (!(ion is IonSymbol
+                        && !ion.isNullValue
+                        && validValues.contains(ion.stringValue()))) {
+            throw InvalidSchemaException("Invalid float_size constraint: $ion")
+        }
+
+        // float16 not implemented yet
+        if (ion.stringValue() == "float16") throw InvalidSchemaException("\'float16\' is not currently supported")
+    }
+
+
+    override fun validate(value: IonValue, issues: Violations) {
+        validateAs<IonFloat>(value, issues) {
+            when (ion.toString()) {
+                "float16" -> throw InvalidSchemaException("\'float16\' is not currently supported")
+                "float32" -> if (!fitsFloat32(it)) {
+                    issues.add(Violation(ion, "insufficient_float_size", "value does not fit float32 size"))
+                }
+                "float64" -> Unit
+            }
+        }
+    }
+
+
+    private fun fitsFloat32(value: IonFloat): Boolean {
+        if (value.isNumericValue) {
+            val absValue = value.bigDecimalValue().abs()
+            if (absValue != BigDecimal.ZERO) {
+                return (maxFloat32Range >= absValue)
+                        && (minPosValue <= absValue)
+            }
+        }
+        return true
+    }
+}
--- a/src/com/amazon/ionschema/internal/constraint/FloatSize.kt
+++ b/src/com/amazon/ionschema/internal/constraint/FloatSize.kt
@@ -11,6 +11,9 @@ import java.math.BigDecimal
 
 /**
  * Implements the float_size constraint.
+ *
+ * float32 validation relies on the precision of constants Float.MAX_VALUE.toDouble() and Float.MIN_VALUE.toDouble().
+ * Client code working with floating points should use .toDouble()'s precision for FloatSize validation to work.
  */
 internal class FloatSize(
         ion: IonValue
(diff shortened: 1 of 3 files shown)


src/com/amazon/ionschema/internal/constraint/FloatSize.kt
```diff
@@ -0,0 +1,59 @@
...
+                "float32" -> if (!fitsFloat32(it)) {
+                    issues.add(Violation(ion, "insufficient_float_size", "value does not fit float32 size"))
+                }
+                "float64" -> Unit
+            }
+        }
+    }
+
+
+    private fun fitsFloat32(value: IonFloat): Boolean {
```
zslayton, line 49: This function already requires that `value` be an `IonFloat`. With this in mind:
* Do we still need to transform the value into a `BigDecimal` before comparing? Or can we get away with the substantially cheaper test of seeing whether `value.doubleValue()` can be stored in a `float` and then cast back to a `double` without losing data?
* Is the `isNumericValue` test required?

src/com/amazon/ionschema/internal/constraint/FloatSize.kt
```diff
@@ -0,0 +1,59 @@
...
+import java.math.BigDecimal
+
+/**
+ * Implements the float_size constraint.
+ */
+internal class FloatSize(
+        ion: IonValue
+) : ConstraintBase(ion) {
+
+    private val validValues = setOf("float16", "float32", "float64")
```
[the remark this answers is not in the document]
therapon: We have had no requests for `float16`. It was added anticipating that we might given some of the use cases we have had where ISL is used to map (or express) types and their constraints from database systems (MySQL etc.)

src/com/amazon/ionschema/internal/constraint/FloatSize.kt
therapon: I think we get the same behaviour with the `double` approach. The reason for the use of `BigDecimal` was to avoid some of the rounding policies that Java has on float and double. 

The `isNumeric` it to distinguish the special float values `NaN` ``-Inf` `Inf` which should fit regardless of the size constraints on `float`

src/com/amazon/ionschema/internal/constraint/FloatSize.kt
```diff
@@ -0,0 +1,59 @@
...
+                    issues.add(Violation(ion, "insufficient_float_size", "value does not fit float32 size"))
+                }
+                "float64" -> Unit
+            }
+        }
+    }
+
+
+    private fun fitsFloat32(value: IonFloat): Boolean {
+        if (value.isNumericValue) {
```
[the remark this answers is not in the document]
therapon: This is to cover the special `float` values `NaN` etc. A comment here, maybe even alter the `if` test to check for the non-numerics first and use the `else` for the numeric values might even be better?

src/com/amazon/ionschema/internal/util/RangeIonNumber.kt
```diff
@@ -35,7 +35,11 @@ internal class RangeIonNumber private constructor (
                 } else {
                     when (ion) {
                         is IonDecimal -> ion.bigDecimalValue()
-                        is IonFloat -> ion.bigDecimalValue()
+                        is IonFloat -> if (ion.isNumericValue) {
```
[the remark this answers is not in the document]
desaikd: For special values `+/- inf, nan`, `isNumericValue` is `false` and range doesn't include these special float values, they need to be added explicitly.

src/com/amazon/ionschema/internal/constraint/FloatSize.kt
zslayton: > I think we get the same behaviour with the double approach. The reason for the use of BigDecimal was to avoid some of the rounding policies that Java has on float and double.

It would be good to add a comment explaining the exact behavior we need, maybe with a link to something explaining the rounding policies you mention. Without that context it's tempting to try to optimize it by avoiding the conversion.

> The isNumeric it to distinguish the special float values NaN ``-Inf Inf` which should fit regardless of the size constraints on `float`

Ahh, got it! I thought that `isNumericValue` was on the `IonValue` interface (to tell if the value is an `int`, `float`, or `decimal`), but I see now that it's actually [a method on `IonFloat`](https://www.javadoc.io/doc/com.amazon.ion/ion-java/latest/com/amazon/ion/IonFloat.html#isNumericValue--). That might also be nice to explain in a comment.

src/com/amazon/ionschema/internal/util/RangeIonNumber.kt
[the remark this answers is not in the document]
zslayton: Thank you (and Theo) for explaining. I wish we'd provided the method `isSpecialValue()` instead of `isNumericValue()`. It would be consistent with [phrasing from the spec](https://amzn.github.io/ion-docs/docs/float.html#special-values) and would avoid the overloaded phrase "numeric." Oh well.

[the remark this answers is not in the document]
zslayton: Could you add a clarifying comment to this branch? Something like:
```suggestion
                        is IonFloat -> if (ion.isNumericValue) {
                            // nan, +inf, -inf
```

src/com/amazon/ionschema/internal/constraint/FloatSize.kt
therapon: Would a comment like the following do? 

```
/**  
 * Implements the float_size  *
                  * float32 validation relies on the precision of constants Float.MAX_VALUE.toDouble() and Float.MIN_VALUE.toDouble().
                  * Client code working with floating points should use .toDouble()'s precision for FloatSize validation to work.
                  */


```

zslayton: That comment makes sense. I guess I was hoping to see an example of a place where `double`->`float`->`double` would lose precision in a way that keeps it from being a good test for `fitsInFloat32`.

therapon: This has an example, no exactly the order in your comment. 
https://stackoverflow.com/questions/916081/convert-float-to-double-without-losing-precision

therapon (approved):

zslayton (approved):

popematt:
Fixed https://github.com/amzn/ion-schema/issues/18

Issue #10, opened 2018-10-26
define a schema for ISL


pbcornell:
This would not only be a good exercise, but also a comforting indicator of completeness ("yes, ISL can be used to define itself").  Once defined, it could be used to validate other ISL.

PR #166, opened 2021-02-22
added changes for negation constraint on timestamp offset

desaikd:
*Description of changes:*
This is PR adds negation constraint on timestamp offset. 

*Test:*
https://github.com/amzn/ion-schema-tests/pull/10

### DIFF
--- a/src/com/amazon/ionschema/internal/constraint/TimestampOffset.kt
+++ b/src/com/amazon/ionschema/internal/constraint/TimestampOffset.kt
@@ -43,6 +43,7 @@ internal class TimestampOffset(
      * exactly with values returned by IonTimestamp.localOffset.
      */
     private val validOffsets: Set<Int?>
+    private val negateOffsets: Boolean
 
     init {
         if (ion !is IonList) {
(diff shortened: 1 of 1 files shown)


therapon (approved):

src/com/amazon/ionschema/internal/constraint/TimestampOffset.kt
```diff
@@ -43,6 +43,7 @@ internal class TimestampOffset(
      * exactly with values returned by IonTimestamp.localOffset.
      */
     private val validOffsets: Set<Int?>
+    private val negateOffsets: Boolean
```
zslayton, line 46: Because the offsets are signed integers, the word `negate` is a bit misleading. How about `offsetsAreValid`, indicating the contents of the `offsets` variable above?

src/com/amazon/ionschema/internal/constraint/TimestampOffset.kt
```diff
@@ -43,6 +43,7 @@ internal class TimestampOffset(
      * exactly with values returned by IonTimestamp.localOffset.
      */
     private val validOffsets: Set<Int?>
```
zslayton, line 45: Now that there's a `not::` option, we don't know whether this `Set` is storing `validOffsets` or `invalidOffsets`. I suggest changing it to just `offsets`.

zslayton (approved):
Couple of naming quibbles, but otherwise looks good.

PR #167, opened 2021-03-31
Update to Gradle 6 and prevent building with the kotlin 1.5.0 preview releases

popematt:
**Issue #, if available:**

None.

**Description of changes:**
Upgrading to Gradle 6 (which is a requirement if we ever upgrade to Kotlin 1.4+). These changes are entirely auto-generated by running `gradle wrapper --gradle-version 6.8.3`, and happened incidentally as I was sorting out the kotlin version issue.

Disallowing the use of Kotlin 1.5 (for now) because the kotlin compiler will start throwing errors if you try to compile pre-release code without using the equivalent pre-release compiler. If anyone actually needs to compile this library with a pre-release version of kotlin, they can still check out this package and locally change the kotlin compiler gradle plugin version and the kotlin std lib version to 1.5.0-M2.
If we go ahead with making this change, I will create a follow-up issue to revisit kotlin 1.5 once it has a stable release.


build.gradle
```diff
@@ -22,7 +22,7 @@ plugins {
 
 group = "com.amazon.ion"
 version = "1.2.0-SNAPSHOT"
-ext.kotlin_version = "[1.3,)"
+ext.kotlin_version = "[1.3,1.5["
```
almann, line 25: This made me lookup what `[` [meant][1] at the end of the range, which leads me to ask is there a reason to prefer `[` over `)` for exclusive end of range or did I miss a difference for this syntax? 

[1]: https://docs.gradle.org/current/userguide/single_versions.html

gradlew
```diff
@@ -28,7 +44,7 @@ APP_NAME="Gradle"
 APP_BASE_NAME=`basename "$0"`
 
 # Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-DEFAULT_JVM_OPTS=""
+DEFAULT_JVM_OPTS='"-Xmx64m" "-Xms64m"'
```
almann, line 47: Was this generated as part of the upgrade, or did you add it?

almann (approved):
:rocket: LGTM--very minor questions below.

gradlew
popematt: This was generated.

build.gradle
popematt: There is no difference between them, and I didn't put any thought into one syntax vs the other. Do we care enough to change it back to `)`? (I'll do it if someone prefers it that way.)

PR #170, opened 2021-06-16
Fixes a bug in IonSchemaTestRunner that caused it to silently ignore some test cases

popematt:
**Issue #, if available:**
N/A

**Description of changes:**

Given a [schema definition test](https://github.com/amzn/ion-schema-tests#user-content-schema-definition-test) like this:

```
schema_header::{}
type::{
  name: Foo,
  type: struct,
  fields: {
    a: int,
    b: string,
  }
}
schema_footer::{}

valid::{
  Foo: [
    { a: 1, b: "hello" },
    { a: 2, b: "world", c: "open content" },
  ],
}

invalid::{
  Foo: [
    { a: "should be int", b: "a string" },
    { a: 1, b: a_symbol },
  ],
}
```
The `IonSchemaTestRunner` would discover the schema header, and pass the sequence of ion values to the `SchemaImpl` constructor. Because ISL itself allows open content, the `SchemaImpl` would consume _all_ the ion values from the test file—including the `valid::` and `invalid::` test cases. Then the test runner tries to run the test cases on the schema, but the sequence of ion values has already been drained, so there is nothing left for it to run.

My change fixes `IonSchemaTestRunner` so that it will detect if a `schema_header` is present in a test file, and if there is, then the `IonSchemaTestRunner` will split the sequence of ion values into "schema" and "test inputs", passing the schema ion to the schema constructor, and retaining the test inputs for running the test cases.

### DIFF
--- a/test/com/amazon/ionschema/AbstractTestRunner.kt
+++ b/test/com/amazon/ionschema/AbstractTestRunner.kt
@@ -46,7 +46,7 @@ abstract class AbstractTestRunner(
             test()
         } catch (ae: AssertionError) {
             notifier.fireTestFailure(Failure(desc, ae))
-        } catch (e: Exception) {
+        } catch (e: Throwable) {
             notifier.fireTestFailure(Failure(desc, e))
         } finally {
             notifier.fireTestFinished(desc)
--- a/test/com/amazon/ionschema/IonSchemaTestRunner.kt
+++ b/test/com/amazon/ionschema/IonSchemaTestRunner.kt
@@ -67,17 +67,23 @@ class IonSchemaTestRunner(
                 var schema: Schema? = null
                 var type: Type? = null
 
-                val iter = ION.iterate(FileReader(file)).asSequence().toList().listIterator()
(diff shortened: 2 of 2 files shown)


popematt:
Failing test cases are the following–which are the tests for allowing ranges inside of a valid_values list.

```
constraints/valid_values/validation_list_timstamp_ranges. 0001-01-01T00:00Z
constraints/valid_values/validation_list_timstamp_ranges. 2000-01-01T00:00Z
constraints/valid_values/validation_list_timstamp_ranges. 2000-02-01T00:00Z
constraints/valid_values/validation_list_timstamp_ranges. 2000-03-01T00:00Z
constraints/valid_values/validation_ranges. -10.
constraints/valid_values/validation_ranges. -15.
constraints/valid_values/validation_ranges. -20.
constraints/valid_values/validation_ranges. 10.
constraints/valid_values/validation_ranges. 15.
constraints/valid_values/validation_ranges. 20.
```

Also, if you have any suggestions for a better way to submit a PR along with corresponding changes in a git submodule, I'm all ears.

desaikd (approved):
LGTM!

Issue #122, opened 2019-06-10
provide a means for making the ISL for ISL easily usable


pbcornell:
Now that #10 is resolved (see the ISL in the `schema/isl` path), we should provide a way for that ISL to be used by customers.  At minimum, this likely means moving the files to a different repository (as they are not specific to the Kotlin implementation), and perhaps providing an API that loads and makes the types defined therein available at runtime.

pbcornell:
The "ISL for ISL" files (formerly in the `schema/isl` path) have been moved to a new [ion-schema-schemas](https://github.com/amzn/ion-schema-schemas) repository, and ion-schema-kotlin now "mounts" that repository as a submodule.

Customers are now free to do the same without taking a dependency on the ion-schema-kotlin repository.  Configuring an instance of AuthorityFilesystem with a base path of `<ion-schema-schemas>/isl` will allow resolution of schema ids such as `isl/schema.isl`.

To avoid bloat (keeping embedded systems use cases in mind), we've intentionally avoided bundling these files with the ion-schema-kotlin distribution, and we've intentionally avoided providing an API specific to the "ISL for ISL" validation use case for similar reasons.  We believe the existing API supports such use cases reasonably well, and will revisit this decision in the future if there are sufficient reasons for doing so.

PR #173, opened 2021-06-21
Adds ion-schema-schemas to the ion-schema-kotlin jar and provides an Authority for accessing them

### LINKED ISSUE #156: Add this ResourceAuthority as part of the public API
Loading resources embedded within a `.jar` seems like a somewhat common thing to have to do.  Please include this (or a variation of it) as part of the public API for `ion-schema-kotlin`.

```Kotlin
class ResourceAuthority(
    val rootPackage: String,
    val classLoader: ClassLoader,
    val ion: IonSystem
) : Authority {

    override fun iteratorFor(iss: IonSchemaSystem, id: String): CloseableIterator<IonValue> {
        val resourceName = "${rootPackage}/$id"
        var str: InputStream? = classLoader.getResourceAsStream(resourceName)
                ?: error("Failed to load schema with resource name '$resourceName'")

        return object : CloseableIterator<IonValue> {

            private var stream = str
            private var reader = ion.newReader(stream).also { it.next() }
            private var iter = ion.iterate(reader)

            override fun hasNext() = iter.hasNext()
            override fun next() = iter.next()
            override fun close() {
                try {
                    reader?.close()
                    stream?.close()
                } finally {
                    reader = null
                    stream = null
                }
            }
        }
    }
}
```

popematt:
**Issue #, if available:**
Fixes #156 

**Description of changes:**
Copies the `ion-schema-schemas` definition of ISL into the `ion-schema-kotlin` build output (as a resource in the jar) so that users of `ion-schema-kotlin` don't need to pull in any other dependency to get the ISL for ISL.

Also adds an Authority for accessing the schemas. The `ResourceAuthority` can load schemas that are bundled as a resource in a jar, and it comes with a static factory method for accessing the ISL for ISL schemas.

N/A

Note that this goes against the decision in #122 . The rationale for ignoring that precedent now is:
1. The extra size is trivial—the total size of the `ion-schema-kotlin` jar is approximately 266kb (~500kb uncompressed), whereas the ISL schemas (uncompressed and using the ion text encoding) are under 9kb. This represents approximately a 2% increase in the size of the build artifact.
2. The ease-of-use improvement is significant for users of the library. These schemas can be used for static analysis at build time, and we should encourage Ion Schema users to do so as a best practice. It's much easier for users if the schemas are bundled in the jar rather than requiring our users to mount a git submodule.

### DIFF
--- a/test/com/amazon/ionschema/ResourceAuthorityTest.kt
+++ b/test/com/amazon/ionschema/ResourceAuthorityTest.kt
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.ionschema
+
+import com.amazon.ion.system.IonSystemBuilder
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertTrue
+import org.junit.Assert.fail
+import org.junit.Test
+
+class ResourceAuthorityTest {
+
+    @Test
+    fun unknownSchemaId() {
+        val iss = IonSchemaSystemBuilder.standard().build()
+        val authority = ResourceAuthority("ion-schema-schemas", ResourceAuthority::class.java.classLoader)
+        val iter = authority.iteratorFor(iss, "unknown_schema_id")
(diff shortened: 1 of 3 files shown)


test/com/amazon/ionschema/ResourceAuthorityTest.kt
```diff
@@ -0,0 +1,61 @@
...
+    fun unknownSchemaId() {
+        val iss = IonSchemaSystemBuilder.standard().build()
+        val authority = ResourceAuthority("ion-schema-schemas", ResourceAuthority::class.java.classLoader)
+        val iter = authority.iteratorFor(iss, "unknown_schema_id")
+        assertFalse(iter.hasNext())
+        try {
+            iter.next()
+            fail()
+        } catch (e: NoSuchElementException) {
+        }
```
jobarr-amzn, line 37: This sent me on a distraction reading about `try-with-resources` support in Kotlin (or lack thereof), the first ultimately sad experience I've had with the language. 

There is [`use`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html) though.

jobarr-amzn: Also it would be great to hop to JUnit5 and use `assertThrows` here.

jobarr-amzn (approved):

Issue #178, opened 2021-10-15
Stop imports of schemas from including transitive dependencies


popematt:
`ion-schema-kotlin` needs to stop allowing transitive imports and pass [these tests](https://github.com/amzn/ion-schema-tests/pull/14/files
) in order to be compliant with the Ion Schema Specification.

This is a backwards incompatible change for certain schemas that have been relying on the current, incorrect behavior that allows transitive imports, so we should also provide a utility to automatically fix affected schemas.

PR #180, opened 2021-11-03
Adds option to choose spec-compliant behavior for resolving schema imports

popematt:
**Issue #, if available:**

Partial fix of #178 

**Description of changes:**

Key Changes:
* Adds a builder option to enable spec compliant behavior for resolving imported types (ie. you cannot get a type transitively through another import)
* Adds a builder option to specify a logging callback
* Logs a warning when attempting to import a type transitively (but it does not cover all possible cases).
* Logs a warning when `validate` is called on any transitively imported type
* Moves `ion-schema-kotlin` to the latest commit of `ion-schema-tests`
* Introduces new functions to `Schema`—`getDeclaredType(String)` and `getDeclaredTypes()`—that can only return types which are declared in the schema (rather than being imported).

Minor/Incidental Changes:
* Adds `ionSystem` as a public property of the `IonSchemaSystem` interface.
* Adds an interface `ImportedType` to help distinguish between imported and declared types of a schema (and to add useful information about where it was imported from).
* Adds a new Authority implementation—`InMemoryMapAuthority`—that is an in-memory map of `schemaId` to `String`/`IonValue`.
* Changes the `params` of `IonSchemaSystemImpl` to something a little nicer—now it has (some) type-safety and the default values are included in the parameter definition rather than at the site where it's used.
* Logs a warning when a schema has no declared types

**Things to pay extra attention when reviewing**
- Are the warning messages clear, useful, etc?
- Is this going to generate too many logs if we warn every time a transitively imported type is used for validation? (I went with this approach because it's considerably simpler to implement than inspecting the whole schema at load-time to find usages, but it _should_ be possible to only warn about usages when a schema is loaded if the current approach would generate too many log messages.)

- https://github.com/amzn/ion-schema/pull/52
- https://github.com/amzn/ion-schema-tests/pull/14


src/com/amazon/ionschema/internal/SchemaImpl.kt
```diff
@@ -154,21 +165,49 @@ internal class SchemaImpl private constructor(
...
-                        var newType = importedSchema.getType(typeName)
-                            ?: throw InvalidSchemaException(
-                                "Schema $childImportId doesn't contain a type named '$typeName'"
-                            )
+                        var importedType = importedSchema.getDeclaredType(typeName)
+
+                        if (importedType == null && allowTransitiveImports) {
+                            importedType = importedSchema.getType(typeName)?.also { type ->
+                                schemaSystem.logger(Warn) {
+                                    "Import is resolved transitively; $typeName is actually declared in " +
```
desaikd, line 173: Instead of saying `Import is resolved transitively` I feel if it is something similar to the option `ALLOW_TRANSITIVE_IMPORTS` then its easier to understand.

```kotlin
"Schema system allows transitive imports; $typeName is actually declared in schema '${type.schemaId}' -- $it' "
```

desaikd (commented):
Overall I think this looks good, although it is true that there are too many log messages. If we could do something like a single warning message for each `*` import instead of for each transitively imported type then it might be enough.

src/com/amazon/ionschema/internal/IonSchemaSystemImpl.kt
```diff
@@ -62,21 +63,24 @@ internal class IonSchemaSystemImpl(
...
 
-    internal enum class Param {
-        ALLOW_ANONYMOUS_TOP_LEVEL_TYPES, // for backwards compatibility with v1.0
+    internal sealed class Param<T : Any>(val defaultValue: T) {
+        // for backwards compatibility with v1.0
+        // Default is to NOT support the backwards compatible behavior
+        object ALLOW_ANONYMOUS_TOP_LEVEL_TYPES : Param<Boolean>(false)
+        // for backwards compatibility with v1.1
+        // Default is to keep the backward compatible behavior
+        object ALLOW_TRANSITIVE_IMPORTS : Param<Boolean>(true)
```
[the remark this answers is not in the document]
popematt: Default will be true for the 1.2 release, and will be changed to false in the 2.0 release.

src/com/amazon/ionschema/Schema.kt
```diff
@@ -54,16 +54,28 @@ interface Schema {
...
-     * Returns the requested type, if present in this schema;
-     * otherwise returns null.
+     * Returns the requested type, if present in this schema, imported
+     * into this schema, or a built-in type; otherwise returns null.
      */
     fun getType(name: String): Type?
 
     /**
-     * Returns an iterator over the types in this schema.
+     * Returns an iterator over all the types declared in this schema or
```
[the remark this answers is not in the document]
popematt: No, actually. I don't know for sure what the intention of the function it, but the actual behavior of the function is to only return imported and declared types, and not any of the built-in types.

src/com/amazon/ionschema/internal/SchemaImpl.kt
popematt: In principle, I like this idea, but think the focus of the message should be on the problem that this particular type is imported in a way that is non-compliant with the spec rather than focusing on the fact that the schema system has transitive imports enabled.

What if there was a warning "code" like `WARN_TRANSITIVE_IMPORT` that for all of the logs related to this? It would be very easily searchable because it's all one word (w.r.t. search engines).

Eg.

```
"WARN_TRANSITIVE_IMPORT; $typeName is actually declared in schema '${type.schemaId}' -- $it' "
```

src/com/amazon/ionschema/IonSchemaSystemLogger.kt
```diff
@@ -0,0 +1,41 @@
...
+ * This allows the IonSchemaSystem implementation to be agnostic to
+ * the actual logging framework used in an application.
+ */
+typealias IonSchemaSystemLogger = (level: LogLevel, message: () -> String) -> Unit
+
+enum class LogLevel {
+    /**
+     * Used for logging general information about the execution of library routines.
+     */
+    Info,
```
tgregg, line 30: It's hard for me to imagine a scenario where we'd actually use the Info level. We don't want to give users a bunch of information about the internal execution of the library; we really only want to give them warnings that indicate they should take an action, which fits the Warn level. And as you say in the documentation of the Warn level, there's no level above that because the library would just throw an error in that case.

I recommend removing the concept of LogLevel.

src/com/amazon/ionschema/IonSchemaSystemBuilder.kt
```diff
@@ -93,7 +95,28 @@ class IonSchemaSystemBuilder private constructor() {
...
+        params[IonSchemaSystemImpl.Param.ALLOW_TRANSITIVE_IMPORTS] = boolean
+        return this
+    }
+
+    /**
+     * Provides a callback for the IonSchemaSystem to use for writing log messages.
+     *
+     * @since 1.2
+     */
+    fun withLogger(logger: IonSchemaSystemLogger): IonSchemaSystemBuilder {
```
tgregg, line 118: The use of the term "log" and "logger" throughout this PR feels too specific to me. I'd rather just call it a callback and state in the documentation that one of the use cases for the callback is to allow the user to log the message. But all this is really doing is allowing the user to receive a string message from the library that something bad (but currently non-exceptional) happened. The user doesn't have to log this; they can ignore it, throw their own error, etc.

src/com/amazon/ionschema/internal/IonSchemaSystemLoggerInternal.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.amazon.ionschema.IonSchemaSystemLogger
+import com.amazon.ionschema.LogLevel
+
+/**
+ * Logger that adds "[ion-schema-kotlin]" prefix to all lines and then delegates to
+ * a logging callback.
+ */
+internal class IonSchemaSystemLoggerInternal(private val delegate: IonSchemaSystemLogger) {
+    operator fun invoke(level: LogLevel, message: () -> String) {
+        delegate.invoke(level) { "[ion-schema-kotlin] " + message() }
```
tgregg, line 27: I'm on the fence about whether this is valuable. The callback is specific to this library, so users who want this could add the prefix themselves. Or dump a stack trace that identifies the specific classes, or do any number of other things.

tgregg (commented):
I don't have concerns about the fix, just a couple of comments about the callback stuff.

> Is this going to generate too many logs if we warn every time a transitively imported type is used for validation?

I'm not too worried about this. The callback gives users complete freedom to do things like:
* Only log the first occurrence
* Only log every 100th occurrence
* Etc.

src/com/amazon/ionschema/internal/WarningMessagesUtils.kt
```diff
@@ -0,0 +1,61 @@
...
+     * IonSchemaSystem would issue this warning if it detects that schema A
+     * references a type that was declared in schema C.
+     *
+     * The import resolution behavior in `ion-schema-kotlin` that allowed this
+     * to happen is incorrect and will be deprecated in future versions
+     * of `ion-schema-kotlin`.
+     *
+     * You can disable this incorrect import resolution behavior now by
+     * setting [IonSchemaSystemBuilder.allowTransitiveImports] to `false`,
+     * but be aware that you should carefully test
```
tgregg, line 26: Incomplete sentence

src/com/amazon/ionschema/IonSchemaSystemBuilder.kt
```diff
@@ -100,23 +101,73 @@ class IonSchemaSystemBuilder private constructor() {
...
+     * Clients can use these warnings as they please. Some possible uses are:
+     * - Log the warning messages
+     * - Surface the warnings to an end user who is authoring schemas
+     * - Enforce a "strict mode" by throwing an exception for warnings (ie. like
+     *   `javac -Werror`, but for Ion Schemas)
+     *
+     * NOTE—This method is equivalent to [withWarningMessageCallback], except that
+     * this method accepts a [Consumer]. This method is intended to be used for
+     * calling from Java code because [withWarningMessageCallback] requires a
+     * function that returns [Unit], which may be awkward for Java callers.
```
tgregg, line 140: Is using Consumer awkward for Kotlin callers?

tgregg (approved):

desaikd (approved):

src/com/amazon/ionschema/IonSchemaSystemBuilder.kt
popematt: Using Consumer is not awkward for Kotlin—just more verbose than using a regular kotlin function. Both methods are usable from either language, they're just each un-idiomatic in one language, but clean and idiomatic in the other.

```java
// Java
IonSchemaSystemBuilder.standard()
        .withWarningMessageConsumer(System.out::println) // <-- idiomatic java
        .withWarningMessageCallback(it -> { System.out.println(it); return Unit.INSTANCE; })
```
```kotlin
// Kotlin
IonSchemaSystemBuilder.standard()
        .withWarningMessageConsumer(Consumer { println(it) })
        .withWarningMessageCallback { println(it) }  // <-- idiomatic Kotlin
```
I poked around a little more to see if there was any nice solution, but Java's `void` is not part of Kotlin's `Any`, so while I could create a functional interface that extends `(String) -> Unit` that doesn't solve the problem. This seems to be a rough edge that is specific to `Consumer`, but would not occur for a java `Supplier<T>` or java `Function<T, R>`. [Here](https://discuss.kotlinlang.org/t/java-interop-unit-closures-required-to-return-kotlin-unit/1842) is an interesting discussion of the problem, if you have time to start going down a rabbit hole.

That being said, it seems like it would be better for me to overload the function rather than have two different names as I do now. I'm going to change that.

Issue #95, opened 2019-04-02
address SchemaImpl TBD when $ion_schema_1_0 is encountered


pbcornell:

popematt:
Fixed in #95

PR #189, opened 2021-12-15
Adds check to make sure ISL version marker is valid

popematt:
**Issue #, if available:**

#95 

**Description of changes:**

Throws an `InvalidSchemaException` if the schema contains an ISL version marker for a version other than 1.0.

This is _technically_ a breaking change, but it should not be a _substantive_ breaking change since there are no other ISL versions in existence, and no one should be adding ISL version markers other than `$ion_schema_1_0`.

Why are we making this change? Because otherwise, users could start creating schemas with (for example) `$ion_schema_2_0`, and whenever support for `$ion_schema_2_0` is released, there could be unexpected behavior changes for the user's schema.

Also, [according to the spec](https://amzn.github.io/ion-schema/docs/spec.html#schema-definitions), schemas are required to start specifically with `$ion_schema_1_0`. Some of our users—as well as our own test cases—have schemas that _do not_ start with an ISL version marker. It would be a substantive breaking change if we started to enforce the presence of the version marker. Minimally, though, we can start by enforcing that the ISL version marker is for a version that actually exists.

None

### DIFF
--- a/src/com/amazon/ionschema/internal/SchemaImpl.kt
+++ b/src/com/amazon/ionschema/internal/SchemaImpl.kt
@@ -61,6 +61,10 @@ internal class SchemaImpl private constructor(
 
     private val declaredTypes: Map<String, TypeImpl>
 
+    companion object {
+        private val ISL_VERSION_MARKER = Regex("^\\\$ion_schema_\\d+_\\d+")
+    }
+
     init {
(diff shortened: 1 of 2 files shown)


desaikd (approved):
🚢

tgregg (approved):

PR #192, opened 2022-06-06
Adds AST model for ISL

popematt:
**Issue #, if available:**

N/A

**Description of changes:**

Adds an AST model for Ion Schema.

This PR is intended for initial feedback. There are some examples that are not intended to be merged as is (the mediators and the pseudo-DSL for constructing ISL), and this code admittedly could use a little more thought about the organization. There are also some places where the AST could be improved by using sealed interfaces, but I think I need to update the Kotlin version for that.

Here are a few things I was trying to accomplish:
* Constraints can be identified by ID (essentially the field name, but a little safer since it doesn't rely on hard-coding strings everywhere, and it includes some type information)
* Constraints can also be tested by using `foo is ValidValuesConstraint` (or `foo instanceOf ValidValuesConstraint` in Java), but there is no guarantee that the classes will uniquely identify the constraint.
* The AST classes should (as much as possible) be plain old data objects.
* Prefer composition rather than inheritance. You see this, for example, in `AstVariablyOccuringType` where rather than extending `AstType`, it is composed of an `AstType` member and an occurs range member.
* Where there is redundancy in ISL, express that through convenience functions rather than redundant data structures. For example, the `container_length` constraint can accept a number or a range. I've modeled it as only a range since the a single number `n` is equivalent to `range::[n, n]`.

I've added a few comments denoted by a 📝.

The build is failing because of `ktlint`. This is (somewhat) intentional—I don't care about those failures right now because I don't want this PR to be merged.

N/A

### DIFF
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/model/AstConstraint.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/model/AstConstraint.kt
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazon.ionschema.model
+
+/**
+ * Represents an Ion Schema Language constraint. Implementations must add any necessary properties for modelling the
+ * details of that particular constraint.
+ *
+ * The recursive type bound enforces that [id] has a valueType that is the same type as the implementation of this interface.
+ *
+ * Sometimes we may need to get a ConstraintId in a static way (i.e. we have no instance of this constraint). For this
+ * reason, it is recommended that all concrete implementations of [AstConstraint] include a companion object like this
+ * example from [AllOfConstraint]:
+ * ```
+ * companion object : ConstraintId<AllOfConstraint> by ConstraintId("all_of") {
+ *     @JvmField val ID = this@Companion
+ * }
+ * ```
+ * The companion object implements [ConstraintId] which means that when using this library from Kotlin, the class
+ * itself is like a static [ConstraintId].
+ * ```
+ * val someId: ConstraintId<*> = AllOfConstraint
+ * ```
+ * When using this library from Java, the ConstraintId can be accessed from `AllOfConstraint.ID`
+ * ```
+ * ConstraintId<?> someId = AllOfConstraint.ID;
+ * ```
+ */
+interface AstConstraint<T : AstConstraint<T>> {
+    val id: ConstraintId<T>
+}
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/model/AstSchema.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/model/AstSchema.kt
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazon.ionschema.model
+
+import com.amazon.ionelement.api.IonElement
+
+/**
+ * Represents a complete Ion Schema document. Note—the model does not include a particular ISL version because the AST
+ * is intended to be a version-agnostic representation of ISL.
+ */
+data class AstSchema(
+    val id: String? = null,
+    val header: AstHeader,
+    val content: List<SchemaContent>,
+    val footer: AstFooter,
+) {
+
+    val types get() = content.filterIsInstance<SchemaContent.NamedType>()
+
+    sealed class SchemaContent {
+        /**
+         * Represents a top-level type declaration in a schema document.
+         */
+        data class NamedType(val name: String, val definition: AstType.TypeDefinition) : SchemaContent() {
+            @JvmOverloads constructor(
+                name: String,
+                constraints: Collection<AstConstraint<*>>,
+                userContent: List<Pair<String, IonElement>> = emptyList()
+            ) : this(
+                name, AstType.TypeDefinition(constraints, userContent)
+            )
+        }
+
+        /**
+         * Represents top-level user content (i.e. "open content") in a schema document.
+         */
+        data class UserContent(private val element: IonElement) : SchemaContent(), IonElement by element
+    }
+}
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/model/AstVariablyOccurringType.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/model/AstVariablyOccurringType.kt
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazon.ionschema.model
+
+/**
+ * Represents a type that has an `occurs` value. Note that this includes even
+ * implicit `occurs`, such as in this example:
+ * ```
+ * type::{
+ *   fields: {
+ *     foo: string  // this `string` has an implicit `occurs: optional`
+ *   }
+ * }
+ * ```
+ */
+data class AstVariablyOccurringType(val occurs: DiscreteRange, val type: AstType)
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/model/ConstraintId.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/model/ConstraintId.kt
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazon.ionschema.model
+
+/**
+ * Constraints are uniquely identified by their name (i.e. their ISL keyword). The [valueType] is included so that it is
+ * always possible to safely cast a constraint to the appropriate type, and so that constraint implementations cannot
+ * (easily) use the incorrect name or value type.
+ *
+ * Note—when creating an API, prefer to use [AnyConstraintId] instead of `TypedConstraintId<*>` so that clients of the
+ * API can use members of the [KnownConstraintIds] enum.
+ */
+interface ConstraintId<T : AstConstraint<T>> : AnyConstraintId {
+    override val valueType: Class<out AstConstraint<T>>
+
+    companion object {
+        /**
+         * Factory function for creating instances of [ConstraintId].
+         */
+        @JvmStatic
+        fun <T : AstConstraint<T>> create(constraintName: String, valueType: Class<T>): ConstraintId<T> = ConstraintIdImpl(constraintName, valueType)
+
+        // Kotlin-idiomatic pseudo-constructor. Not callable from Java because of the reified type parameter.
+        inline operator fun <reified T : AstConstraint<T>> invoke(constraintId: String) = create(constraintId, T::class.java)
+    }
+}
+
+private data class ConstraintIdImpl<T : AstConstraint<T>>(override val constraintName: String, override val valueType: Class<T>) : ConstraintId<T>
+
+/**
+ * Use this interface instead of `TypedConstraintId<*>`. This interface exists because it is not allowed for an enum
+ * class to implement a generic interface.
+ *
+ * @see ConstraintId
+ */
+interface AnyConstraintId {
+    val constraintName: String
+    val valueType: Class<out AstConstraint<*>>
+}
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/model/ConstraintMediator.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/model/ConstraintMediator.kt
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazon.ionschema.model
+
+/**
+ * Interactions between the Ion Schema System and [AstConstraint]s are mediated through ConstraintMediators.
+ *
+ * A ConstraintMediator can hold any kind of data associated to [ConstraintId]s, but it is intended primarily for holding
+ * things such as reader, writer, and validator functions where we need preserve some type information that would
+ * normally be lost through type-erasure.
+ */
+abstract class ConstraintMediator<R : ConstraintMediator.Registration<*>>(private val registrations: List<R>) {
+    /**
+     * Tests whether the mediator contains any registration for the given constraint name.
+     */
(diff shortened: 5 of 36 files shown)


ion-schema/src/main/kotlin/com/amazon/ionschema/model/Isl.kt
```diff
@@ -0,0 +1,210 @@
...
+ */
+@file:JvmName("IonSchemaBuilders")
+package com.amazon.ionschema.model
+
+import com.amazon.ionelement.api.IonElement
+import com.amazon.ionelement.api.ionSymbol
+import com.amazon.ionschema.model.constraints.*
+import java.math.BigDecimal
+
+// This file contains a straw-man example of a dsl for creating Ion Schemas.
```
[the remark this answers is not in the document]
alancai98: Similarly, are there plans for any builtin-visitor/rewriting functions? The ISL domain model generated by PIG part of [partiql-isl-kotlin](https://code.amazon.com/packages/Maven-org-partiql_partiql-isl-kotlin) provides some visitors and visitor transforms.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/AstSchema.kt
```diff
@@ -0,0 +1,51 @@
...
+ * permissions and limitations under the License.
+ */
+package com.amazon.ionschema.model
+
+import com.amazon.ionelement.api.IonElement
+
+/**
+ * Represents a complete Ion Schema document. Note—the model does not include a particular ISL version because the AST
+ * is intended to be a version-agnostic representation of ISL.
+ */
```
alancai98, line 22: So how can you be sure the AST will be version-agnostic? E.g. what if a future ISL version were to rename a constraint?

alancai98 (commented):
From the initial review and offline discussion we had, I think this is a reasonable starting point for an AST model for ISL and supports the most of the use cases I had in mind for the PartiQL [dataguide/schema discovery](https://github.com/partiql/partiql-lang-kotlin/tree/3715b393e68e7b9c54f640567ee22c282544eaea/lang/src/org/partiql/lang/schemadiscovery) work. Left some comments related to how some constraints were modeled (e.g. numeric ranges, occurs) and some questions about how the APIs will evolve and support other use cases.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/ConstraintId.kt
```diff
@@ -0,0 +1,51 @@
...
+ * permissions and limitations under the License.
+ */
+package com.amazon.ionschema.model
+
+/**
+ * Constraints are uniquely identified by their name (i.e. their ISL keyword). The [valueType] is included so that it is
+ * always possible to safely cast a constraint to the appropriate type, and so that constraint implementations cannot
+ * (easily) use the incorrect name or value type.
+ *
+ * Note—when creating an API, prefer to use [AnyConstraintId] instead of `TypedConstraintId<*>` so that clients of the
```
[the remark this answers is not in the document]
popematt: `TypedConstraintId<*>` is a name that got changed as I was working on this. I guess I missed this comment. It should say something like this:
```suggestion
 * When creating APIs that accept `ConstraintId<*>` (specifically the wildcard typed `ConstraintId`), use this interface instead.
 * This interface exists because it is not allowed for an enum class to implement a generic interface, and so this allows [KnownConstraintIds] to implement `AnyConstraintId`.
```

ion-schema/src/main/kotlin/com/amazon/ionschema/model/Isl.kt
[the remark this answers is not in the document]
popematt: The file `ConstraintSerdeMediator` has some examples of _partial_ serialization to and from Ion. The plan is that the `IonSchemaSystem` will use this model, and so there will be full support for reading and writing multiple versions of Ion Schema Language.

Visitor/rewriting functions are a good idea. I'll see what I can do.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/AstVariablyOccurringType.kt
```diff
@@ -0,0 +1,28 @@
...
+ * implicit `occurs`, such as in this example:
+ * ```
+ * type::{
+ *   fields: {
+ *     foo: string  // this `string` has an implicit `occurs: optional`
+ *   }
+ * }
+ * ```
+ */
+data class AstVariablyOccurringType(val occurs: DiscreteRange, val type: AstType)
```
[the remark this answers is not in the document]
popematt: `optional` == `range::[0, 1]`
`required` == `range::[1, 1]` == `1`

When there are multiple ways of expressing the same thing, I've tried to have only one way of representing it in the AST, and it will be up to the serializers to determine how it gets written out to Ion. I imagine that for `occurs`, if there's a `range::[2, 2]` the serializer will be smart enough to just output `2`, and for `range::[0, 1]` it will output `optional`.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/Isl.kt
```diff
@@ -0,0 +1,210 @@
...
+// Ideally, this would be exposed as java builders and a kotlin type-safe DSL, like
```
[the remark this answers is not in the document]
popematt: Yes, I think so.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/AstSchema.kt
popematt: I am not _sure_ that it always can be, but it is my intention. I do have a few thoughts about how to handle changing/adding/removing constraints:

* The AST does not need to have a 1:1 mapping with any ISL keywords. The goal is accurately representing the function/behavior of a type definition.
* It is possible to model a constraint that has more than one ID with this AST. If ISL 3.0 changes the name of `valid_values` to `ok_values`, we can just create a new ID:  `val OkValues = ConstraintId<ValidValuesConstraint>("ok_values")`.
* Suppose we have a `foo` constraint that is only in ISL 3.0 and a `bar` constraint that is only in ISL 4.0, it would be possible to have both a `FooConstraint` and a `BarConstraint` in the AST. If you include both of them in a schema, the serializer may reject it as being invalid, but it is _possible_ for them to co-exist in this AST.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/constraints/Constraints.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import com.amazon.ionschema.model.AstConstraint
+import com.amazon.ionschema.model.AstType
+import com.amazon.ionschema.model.DiscreteRange
+
+/**
+ * Common interface for constraints that have range over discrete integers as their only argument.
+ * This allows for code reuse when reading, writing, and validating the `scale`, `precision`, and `*_length` constraints.
+ */
+interface DiscreteRangeConstraint<T> : AstConstraint<T> where T : AstConstraint<T>, T : DiscreteRangeConstraint<T> {
```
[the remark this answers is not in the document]
popematt: The only constraint that models ranges of other sorts of numbers is `valid_values`. There is support for _ranges_ of other numbers, just not a common interface for _constraints_ with other types of ranges because there is no commonality to be found (as of ISL 1.0).

ion-schema/src/main/kotlin/com/amazon/ionschema/model/AstSchema.kt
```diff
@@ -0,0 +1,51 @@
...
+data class AstSchema(
+    val id: String? = null,
+    val header: AstHeader,
+    val content: List<SchemaContent>,
+    val footer: AstFooter,
+) {
+
+    val types get() = content.filterIsInstance<SchemaContent.NamedType>()
```
rmarrowstone, line 30: It might be surprising for users that a property access does this filtering. If you were to make it lazy then it would only need to filter once.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/AstSchema.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    sealed class SchemaContent {
+        /**
+         * Represents a top-level type declaration in a schema document.
+         */
+        data class NamedType(val name: String, val definition: AstType.TypeDefinition) : SchemaContent() {
+            @JvmOverloads constructor(
+                name: String,
+                constraints: Collection<AstConstraint<*>>,
+                userContent: List<Pair<String, IonElement>> = emptyList()
```
rmarrowstone, line 40: What is this modeling? Is it intended to be an Ion Struct but defined a different way?

ion-schema/src/main/kotlin/com/amazon/ionschema/model/ConstraintMediator.kt
```diff
@@ -0,0 +1,53 @@
...
+package com.amazon.ionschema.model
+
+/**
+ * Interactions between the Ion Schema System and [AstConstraint]s are mediated through ConstraintMediators.
+ *
+ * A ConstraintMediator can hold any kind of data associated to [ConstraintId]s, but it is intended primarily for holding
+ * things such as reader, writer, and validator functions where we need preserve some type information that would
+ * normally be lost through type-erasure.
+ */
+abstract class ConstraintMediator<R : ConstraintMediator.Registration<*>>(private val registrations: List<R>) {
```
rmarrowstone, line 24: Having a single place to encapsulate the casts that we (the programmers) know to be safe is a good idea. I'm less keen on having that be abstract, with the presumed intention that it be sub-classed as opposed to used by composition.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/AstConstraint.kt
```diff
@@ -0,0 +1,43 @@
...
+ * itself is like a static [ConstraintId].
+ * ```
+ * val someId: ConstraintId<*> = AllOfConstraint
+ * ```
+ * When using this library from Java, the ConstraintId can be accessed from `AllOfConstraint.ID`
+ * ```
+ * ConstraintId<?> someId = AllOfConstraint.ID;
+ * ```
+ */
+interface AstConstraint<T : AstConstraint<T>> {
```
rmarrowstone, line 41: I find the use of `Ast` prefixing for this and others a bit confusing. It's really just the schema as domain model.

I'm also slightly skeptical of the value of having the interface 'have-an' `id` for itself. I see how that enforces what you want, but it comes at the cost of the cognitive load of the self referential type parameter. I think you could just establish a convention instead, I may be wrong.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/Isl.kt
```diff
@@ -0,0 +1,210 @@
...
+fun timestampPrecision(value: TimestampPrecisions) = TimestampPrecisionConstraint(TimestampPrecisionRange(value, value))
+fun timestampOffset(offset: TimestampOffsetConstraint.Offset) = TimestampOffsetConstraint(offset)
+fun timestampOffset(hours: Int, minutes: Int = 0) = TimestampOffsetConstraint(TimestampOffsetConstraint.Offset.Minutes(hours * 60 + minutes))
+fun timestampOffset(minutes: Int) = TimestampOffsetConstraint(TimestampOffsetConstraint.Offset.Minutes(minutes))
+fun type(type: AstType) = TypeConstraint(type)
+fun utf8ByteLength(range: DiscreteRange) = Utf8ByteLengthConstraint(range)
+fun utf8ByteLength(value: Int) = Utf8ByteLengthConstraint(DiscreteRange(value, value))
+fun validValues(values: Iterable<ValidValue>) = ValidValuesConstraint(values)
+fun validValues(vararg values: ValidValue) = ValidValuesConstraint(values.toList())
+fun validValues(vararg ionValues: IonElement) = ValidValuesConstraint(ionValues.map { ValidValue.SingleValue(it) })
```
desaikd, line 75: This should also include functions to construct `valid_values` from a range (timestamp/number)?

ion-schema/src/main/kotlin/com/amazon/ionschema/model/AstVariablyOccurringType.kt
```diff
@@ -0,0 +1,28 @@
...
+ */
+package com.amazon.ionschema.model
+
+/**
+ * Represents a type that has an `occurs` value. Note that this includes even
+ * implicit `occurs`, such as in this example:
+ * ```
+ * type::{
+ *   fields: {
+ *     foo: string  // this `string` has an implicit `occurs: optional`
```
desaikd, line 23: I am not sure if I follow how this implicit `occurs` is added but for `fields` the default is `occurs: optional` and for `ordered_elements` it is `occurs: required`.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/constraints/AnnotationsConstraint.kt
```diff
@@ -0,0 +1,25 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazon.ionschema.model.constraints
+
+import com.amazon.ionschema.model.AstType
+import com.amazon.ionschema.model.ConstraintId
+
+data class AnnotationsConstraint(override val type: AstType) : UnaryTypeConstraint<AnnotationsConstraint> {
```
desaikd, line 20: Shouldn't this take a list of allowed symbols/annotations instead of `AstType`? 
Maybe have a new class representing `Annotation` with `required`,`closed` or `ordered` properties, which could be used here.

popematt: I've chosen to use something that aligns with the proposed syntax in https://github.com/amzn/ion-schema/issues/51, assuming that the proposal will be adopted.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/AstVariablyOccurringType.kt
popematt: In ISL, the `occurs` can be implicit. In this model it's always explicit. It's `val occurs: DiscreteRange` in `AstVariablyOccuringType`.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/AstSchema.kt
popematt: That is true, but then it would also mean allocating a `Lazy` object and duplicating a subset of the original list. It's something worth considering, but I'd rather deal with it later since it's a backwards compatible change in either direction.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/AstSchema.kt
popematt: I'm assuming you are referring specifically to this line:
```kotlin
userContent: List<Pair<String, IonElement>> = emptyList()
```

This is modeling any "open content" in the type definition. Ion Schema allows you to throw extra fields into a type definition, and Ion Schema implementations will accept it but not use it for anything. However, it's recommended that implementations provide some way for users to access the open content in the schema.

I should probably change it to `List<Field>`. Does that sound reasonable?

ion-schema/src/main/kotlin/com/amazon/ionschema/model/ConstraintMediator.kt
popematt: I also prefer composition to inheritance, but when I created this PR, I hadn't figured out a way to make it work in Kotlin to have a function that essentially returns `A<B>` where `A` and `B` are independent type parameters.

Thanks for challenging me on this. I returned to the problem and figured out a different way to model the types that works in Kotlin's type system.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/AstConstraint.kt
popematt: > I find the use of Ast prefixing for this and others a bit confusing. It's really just the schema as domain model.

It is a domain model, but that isn't mutually exclusive with it being an abstract syntax tree when you're writing a programming language interpreter. (That's essentially what an Ion Schema System is.)

However, I agree that the prefix is unnecessary. 

> I'm also slightly skeptical of the value of having the interface 'have-an' id for itself. I see how that enforces what you want, but it comes at the cost of the cognitive load of the self referential type parameter. I think you could just establish a convention instead, I may be wrong.

Are you referring to the static or instance variable?
* The instance variable is necessary because the constraint implementation class is not enough to identify a constraint (that is assuming that we're going to allow users to bring their own custom constraints). Constraints are uniquely identified by their respective ISL keywords. The fact that the `ConstraintId` introduces a self-referential type parameter into `Constraint` is intentional, and I don't think it adds any cognitive load unless you are trying to implement your own constraint. If you are trying to implement a custom constraint, then it does increase the cognitive load, but I think that can be mitigated with better documentation, and I think the trade-off is worth it in order to have stronger guarantees about type safety.
* The static variable is not strictly necessary, but it seemed like a convenient place to put it where someone could find it.  I realize, though, that I was trying to be too clever by making the companion object itself be the ID, so I'm cutting that bit out. I could also get rid of the `@JvmStatic val ID` in favor of having all of the known constraint ids as members of a pseudo-enum `object`. Do you have an opinion on that?

ion-schema/src/main/kotlin/com/amazon/ionschema/model/Isl.kt
popematt: `fun validValues(vararg values: ValidValue)` will get you pretty close already.

```kotlin
validValues(ValidValue.NumberRange(/* ... */))
validValues(ValidValue.TimestampRange(/* ... */))
```
Or, if we added convenience functions for creating these ranges:
```kotlin
validValues(range(BigDecimal.ONE, Max))
```

That being said, this file (`Isl.kt`) is meant as an example to show that a DSL is possible, but I think it would be better to copy what the AWS Kotlin SDK does and provide Java-style builders _and_ Kotlin-style DSL.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/ConstraintSerdeMediator.kt
```diff
@@ -0,0 +1,71 @@
...
+    val write: (T) -> IonElement
+) : ConstraintMediator.Registration<T>
+
+// Example of how the ConstraintMediator can be used to read ISL and construct the AST.
+fun readType(ion: StructElement, serdeMediator: ConstraintSerdeMediator): AstType.TypeDefinition {
+    return AstType.TypeDefinition(
+        constraints = ion.fields
+            .filter { it.name in serdeMediator } // Find the fields that are constraints
+            .map { serdeMediator.readConstraint(it) } // Read the fields as AstConstraint
+            .toSet(),
```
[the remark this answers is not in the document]
popematt: I believe that is a mistake on my part. It should not be a set.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/AstConstraint.kt
rmarrowstone: Sorry for the lateness of reply... I also regret that I don't have more time to understand better, so take this with a grain of salt...

I guess what I can't tell is why it's important that the Id tie to the type of Constraint. I posit that's not actually what's important. What I think is important is that, for an identifier you know how to construct an instance of that Constraint. That makes me think of two potential tweaks:
1. Make the ConstraintRegistration the API that implementers implement. That "has-an" Id and "has-a" Constraint.
2. But it doesn't actually "have-a" Constraint! It has the ability to construct a Constraint from the (typed) arguments you expect to specify it with.
IDK, I may be wildly off, but that's what I got rn.

PR #195, opened 2022-06-16
Fix implementation of `ordered_elements`

popematt:
**Issue #, if available:**

https://github.com/amzn/ion-schema/issues/62

**Description of changes:**

Fixes the incorrect behavior described in https://github.com/amzn/ion-schema/issues/62. Also updates the `ion-schema-tests` submodule to the latest commit.

https://github.com/amzn/ion-schema-tests/pull/18

### DIFF
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/constraint/OrderedElementsNfaStatesBuilder.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/constraint/OrderedElementsNfaStatesBuilder.kt
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.ionschema.internal.constraint
+
+import com.amazon.ion.IonValue
+
+/**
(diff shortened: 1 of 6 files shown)


desaikd (approved):
Are we removing the previous state machine with a separate PR?

popematt:
> Are we removing the previous state machine with a separate PR?

I thought I had included that... but it looks like I forgot.

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/constraint/OrderedElementsNfaStatesBuilder.kt
```diff
@@ -0,0 +1,130 @@
...
+ *      : [I:1]
+ *    0 : [S1:1]
+ *    1 : [S2:1, S3:1]  <-- Neither one of these paths have a valid transition for `"Hi!"`,
+ *    2 : []                so they both die off. There are 0 possible paths remaining.
+ *  END : []
+ * ```
+ */
+internal class OrderedElementsNfaStatesBuilder {
+
+    private val stateInputs = mutableListOf<NFA.State<IonValue>>(NFA.State.Initial)
```
tgregg, line 93: I wonder if it might make sense to provide a base NFA builder that handles adding the initial and final states, since this boilerplate (as far as I understand) is required and must be duplicated by every NFA application.

tgregg (commented):
Since `NFA` is general-purpose (i.e. not limited to use in the `ordered_elements` context), it probably makes sense to add some tests for NFA directly. Are there any aspects of NFA that aren't well-covered by your OrderedElementsTest? Cycles?

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/constraint/OrderedElementsNfaStatesBuilder.kt
popematt: I don't think a base NFA builder is necessary.
* There's no requirement that other state builders will model the inputs as a list.
* Using inheritance to get rid of 2-3 lines of trivial code seems like overkill.
* `ion-schema-kotlin` is already a difficult library to read (IMO) because of an overuse of inheritance.

tgregg (approved):

Issue #207, opened 2022-09-28
Add support for Ion Schema 2.0


popematt:
- [x] Constraints
  - [x] Logic constraints: all_of, any_of, one_of, not, type
  - [x] Length constraints: byte_length, codepoint_length, container_length, utf8_byte_length
  - [x] annotations
  - [x] contains
  - [x] element
  - [x] exponent
  - [x] field_names
  - [x] fields
  - [x] ieee754_float
  - [x] ordered_elements
  - [x] precision
  - [x] regex
  - [x] timestamp_offset
  - [x] timestamp_precision
  - [x] valid_values
- [x] Import resolution
- [x] `$null_or` modifier
- [x] Top-level open content
- [x] Open content in headers
- [x] Open content in footers
- [x] Open content in types
- [x] Ion Schema version markers
- [x] Ion Schema 1.0 interop

PR #208, opened 2022-09-28
Adds initial support for multiple Ion Schema versions to exist

popematt:
**Issue #, if available:**

#207

**Description of changes:**

* Updates the tests to use the new `ion-schema-tests` format. Still only using ISL 1.0 tests, though.
  * Almost complete rewrite of Ion Schema Test Runner
  * Updates to many other test files that had dependencies on 
* Adds `IonSchemaVersion` enum
* Updates `Schema` to have an `IonSchemaVersion`
  * Adds changes that are spillovers from this, such as `ConstraintFactory` must now be version aware
  * `SchemaImpl` is renamed to `SchemaImpl_1_0` and I created a new `SchemaImpl` with a pseudo-constructor that will delegate to `SchemaImpl_1_0` or `SchemaImpl_2_0` as needed.
* I also refactored the `loadSchema` function in `IonSchemaSystemImpl` because it was giving cryptic error messages when a schema was invalid. Now it's a little better.


ion-schema/src/main/kotlin/com/amazon/ionschema/internal/ConstraintFactory.kt
```diff
@@ -25,7 +25,7 @@ internal interface ConstraintFactory {
     /**
      * If [name] is a recognized constraint name, returns `true`, otherwise `false`.
      */
-    fun isConstraint(name: String): Boolean
+    fun isConstraint(name: String, schema: Schema): Boolean
```
desaikd, line 28: Can we just pass the Isl version here instead of passing the entire schema?

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/ConstraintFactory.kt
```diff
@@ -25,7 +25,7 @@ internal interface ConstraintFactory {
     /**
      * If [name] is a recognized constraint name, returns `true`, otherwise `false`.
```
desaikd, line 26:
```suggestion
     * If [name] is a recognized constraint name, returns `true`, otherwise `false` based on the ISL version that given [schema] uses.
```

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/ConstraintFactory.kt
popematt: We could. I made an arbitrary decision for `isConstraint` and `constraintFor` to have the same parameters, but I can change it if you think it's worth doing.

desaikd: I see. I don't think its a hard requirement to make this change, it could work either way. But I just wanted to make it so that it only requires the value that is needed for `isConstraint` implementation rather than entire `schema`.

popematt:
I ended up refactoring `ConstraintFactoryDefault` so that we don't need to create a new enum for every version.
I also changed the `IonSchemaVersion` values to be `v1_0`, etc, so that it's not as verbose to use them.

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/ConstraintFactoryDefault.kt
```diff
@@ -40,58 +41,55 @@ import com.amazon.ionschema.internal.constraint.TimestampPrecision
...
-        valid_values({ ion, schema -> ValidValues(ion) }),
+    /**
+     * Represents a mapping of (constraint name + supported ISL versions) to a constructor function
+     */
+    private data class ConstraintConstructor(
+        val name: String,
+        val versions: ClosedRange<IonSchemaVersion>,
+        val newInstance: (ion: IonValue, schema: Schema) -> Constraint,
+    ) {
+        constructor(name: String, versions: ClosedRange<IonSchemaVersion>, newInstance: (IonValue) -> Constraint) : this(name, versions, { ion, _ -> newInstance(ion) })
```
desaikd, line 58: (Question) Is this used to return constraint that has same functionalities for multiple versions like for `v1_0` and `v2_0` ?

popematt: Yes. For example, `all_of` is never reasonably going to change.

desaikd (approved):

PR #210, opened 2022-10-05
Adds SchemaImpl_2_0 with support for a subset of constraints

popematt:
*Issue #, if available:*

#207 

*Description of changes:*

* Adds `SchemaImpl_2_0`. Rather than add a lot of branching into a single `Schema` implementation, I figured it would be more readable and easier to not make a mistake by having two separate implementations. I will not be duplicating everything—only things where there is substantially different logic to write. For now, the difference is that `SchemaImpl_2_0` has much stricter validations that `SchemaImpl_1_0` and it has no branches to support transitive imports. There still needs to be further work to support the `user_content` field in the schema header, and how transitive imports will (not) work with Ion Schema 1.0 schemas that use transitive imports.
* Renames `Annotations.kt` to `Annotations_1_0.kt`. This constraint is the only constraint where I think the logic diverges enough to justify a completely separate implementation.
* Adds `TypeOrNullDecorator` to support `$null_or`
* Adds support for most of the constraints that are basically unchanged from 1.0 to 2.0—logic constraints, length constraints, `contains`, `precision`, `timestamp_offset`, `timestamp_precision`, and `valid_values`.
* Updates `ion-schema-tests` submodule to the latest commit
* Updates the `IonSchemaTestsRunner` to handle `invalid_schemas` tests cases.
* Adds a test class for IonSchema 2.0 tests!
* Something kind of cool in this PR is that I'm using the Kotlin contract API for the first time. They are experimental, so they might need to be updated whenever we next upgrade the Kotlin version, but there's only a few of them, and they make the code more readable in other places by allowing the compiler to infer things about the type of a variable. (Here's [an introduction to Kotlin contracts](https://www.baeldung.com/kotlin/contracts) if you're interested in learning more.)

### DIFF
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/ConstraintFactoryDefault.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/ConstraintFactoryDefault.kt
@@ -18,9 +18,10 @@ package com.amazon.ionschema.internal
 import com.amazon.ion.IonValue
 import com.amazon.ionschema.IonSchemaVersion
 import com.amazon.ionschema.IonSchemaVersion.v1_0
+import com.amazon.ionschema.IonSchemaVersion.v2_0
 import com.amazon.ionschema.Schema
 import com.amazon.ionschema.internal.constraint.AllOf
-import com.amazon.ionschema.internal.constraint.Annotations
+import com.amazon.ionschema.internal.constraint.Annotations_1_0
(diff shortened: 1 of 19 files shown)


zslayton (approved):
The code looks good to me, but as I read through it I noticed how heavily I was relying on my familiarity with the spec to follow the logic. I would appreciate having more human-friendly comments describing what different methods/blocks were aiming to achieve, especially as I'm less familiar with Kotlin than other languages.

PR #219, opened 2022-10-11
Adds support for Ion Schema 2.0 annotations constraint

popematt:
**Issue #, if available:**

#207 

**Description of changes:**

* Adds an implementation of the `annotations` constraint for Ion Schema 2.0
* Updates the `IonSchemaTests_2_0` suite to include the `annotations` test cases

* Test cases created in https://github.com/amzn/ion-schema-tests/pull/30

### DIFF
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/constraint/Annotations_2_0.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/constraint/Annotations_2_0.kt
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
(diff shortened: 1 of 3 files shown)


ion-schema/src/main/kotlin/com/amazon/ionschema/internal/constraint/Annotations_2_0.kt
```diff
@@ -0,0 +1,101 @@
...
+            val annotationSymbols = ion.mapTo(ion.system.newEmptyList()) {
+                islRequireIonTypeNotNull<IonSymbol>(it) { "annotations list values must be non-null symbols" }
+                islRequire(it.typeAnnotations.isEmpty()) { "annotations list values may not be annotated" }
+                it.clone()
+            }
+
+            val annotationListTypeIon = ion.system.newEmptyStruct().apply {
+                if (ion.hasTypeAnnotation("closed")) {
+                    // May only contain...
+                    add("element", ion.system.newEmptyStruct().apply { add("valid_values", annotationSymbols.clone()) })
```
tgregg, line 67: Is the `.clone()` necessary?

tgregg (approved):

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/constraint/Annotations_2_0.kt
popematt: Thanks for pointing this out. This one is not strictly necessary because only 2 out of the 3 `clone()` calls in this `init` block are necessary. I think I'll get rid of the `clone()` on line 61 instead of this one, so that we only clone in the branch where it's actually going to be used.

tgregg (approved):

PR #228, opened 2023-01-25
Adds useful violation messages for ordered_elements

### LINKED ISSUE #110: re-enhance violation details provided by the ordered_elements constraint
As part of full implementation of the ordered_elements constraint, #108 removed some potentially helpful validation detail because the StateMachine may follow multiple solution paths, and if multiple paths fail, it's not clear how to provide such validation detail to the caller.

We could consider re-adding such violation details in cases where there is a single path through the list of ordered elements.

popematt:
**Issue #, if available:**

Fixes #110 

**Description of changes:**

This adds better messages for the ordered elements constraint. It's impossible to fully determine what the "cause" of the validation failure is because that requires a better understanding of the data than what is in scope for Ion Schema, and it seems like it also requires some amount of "intuition".

For example, if you expect letters in alphabetical order, then `(a b d c)` is invalid. The _fact_ that it is invalid will be detected at the letter 'c', but it's not really decidable whether the problem is the 'c' or the 'd' that precedes it.

The strategy I've chosen for the violation messages is to report all violations for the first element that does not match any possible "branches" in the NFA.  See `OrderedElementsTest.kt` for some examples of what this looks like.

In order to make this possible, I refactored the NFA class so that it state transitions can return Violations instead of just `true`/`false`, and I had to update the `OrderedElements` constraint to interpret the new outcome data from the NFA.

None.

### DIFF
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/TypeCore.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/TypeCore.kt
@@ -22,6 +22,7 @@ import com.amazon.ionschema.Violation
 import com.amazon.ionschema.Violations
 import com.amazon.ionschema.internal.constraint.ConstraintBase
 import com.amazon.ionschema.internal.util.markReadOnly
+import com.amazon.ionschema.internal.util.schemaTypeName
 
 /**
  * Instantiated to represent individual Core Types as defined by the
@@ -36,9 +37,7 @@ internal class TypeCore(
         else -> IonType.valueOf(nameSymbol.stringValue().toUpperCase())
     }
 
-    private val ionTypeName = ionType.schemaTypeName()
-
-    override val name = ionTypeName
+    override val name = ionType.schemaTypeName()
 
     override val schemaId: String? = null
 
@@ -54,7 +53,7 @@ internal class TypeCore(
                 Violation(
                     ion, "type_mismatch",
                     "expected type %s, found %s".format(
-                        ionTypeName,
+                        name,
                         value.type.schemaTypeName()
                     )
                 )
@@ -63,9 +62,4 @@ internal class TypeCore(
             issues.add(CommonViolations.NULL_VALUE(ion))
         }
     }
-}
-
-internal fun IonType.schemaTypeName() = when (this) {
-    IonType.DATAGRAM -> "document"
-    else -> this.toString().toLowerCase()
 }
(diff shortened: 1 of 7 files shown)


ion-schema/src/main/kotlin/com/amazon/ionschema/internal/TypeCore.kt
```diff
@@ -64,8 +63,3 @@ internal class TypeCore(
         }
     }
 }
-
-internal fun IonType.schemaTypeName() = when (this) {
-    IonType.DATAGRAM -> "document"
-    else -> this.toString().toLowerCase()
-}
```
popematt, line 71: This function was moved to `IonValueUtils.kt` since I'm also using it in `OrderedElements.kt` now.

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/TypeCore.kt
```diff
@@ -36,9 +37,7 @@ internal class TypeCore(
         else -> IonType.valueOf(nameSymbol.stringValue().toUpperCase())
     }
 
-    private val ionTypeName = ionType.schemaTypeName()
```
popematt, line 39: This change is incidental. I noticed that the `ionTypeName` variable is completely redundant (it's a private variable that is always the same as `name`), so I removed it.

popematt:
I'm not sure why the GitHub actions workflow isn't working for the PR. For now, you can verify that it builds successfully by looking at the workflow that ran on push [here](https://github.com/popematt/ion-schema-kotlin/commit/4e7f0c756bd50d796377d874cbc2fa7fddf8061f). I'll look into fixing the workflow for PRs.

alancai98:
> I'm not sure why the GitHub actions workflow isn't working for the PR. For now, you can verify that it builds successfully by looking at the workflow that ran on push [here](https://github.com/popematt/ion-schema-kotlin/commit/4e7f0c756bd50d796377d874cbc2fa7fddf8061f). I'll look into fixing the workflow for PRs.

Vaguely remember this happening before for some PartiQL repos. Think this comment is relevant for why the GH Actions workflows aren't running: https://github.com/amazon-ion/ion-schema-kotlin/pull/202#issuecomment-1184127214

ion-schema/src/test/kotlin/com/amazon/ionschema/internal/constraint/OrderedElementsTest.kt
```diff
@@ -35,6 +35,19 @@ class OrderedElementsTest {
...
+    @MethodSource("messageTestCases")
+    fun test_messages(_name: String, constraintIon: String, value: String, message: String) {
+        val constraint = OrderedElements(ION.singleValue(constraintIon), ISS.newSchema())
+        val v = Violations()
+        constraint.validate(ION.singleValue(value), v, debug = true)
+        // Trim all whitespace to normalize before comparing
+        val actual = v.toString().trim()
+        val expected = message.trim()
+        // We add a newline to make the assertion message easier to read
+        assertEquals("\n$expected\n", "\n$actual\n")
```
jobarr-amzn, line 48: Thanks! :D

jobarr-amzn (approved):

PR #231, opened 2023-02-10
Multiple clean up tasks

### LINKED ISSUE #215: Try to avoid regenerating instances of `Occurs` every time `Fields.validate()` is called.
During initialization, the fields constraint creates a map of field names to "Occurs" values, and then discards it.

https://github.com/amzn/ion-schema-kotlin/blob/d074c290855f4f538ec3740b646757015368cfc0/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/constraint/Fields.kt#L53-L56

But then it looks like almost the same "Occurs" values get regenerated every time that `validate()` is called.

https://github.com/amzn/ion-schema-kotlin/blob/d074c290855f4f538ec3740b646757015368cfc0/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/constraint/Fields.kt#L65-L73

It seems like it should be possible to store these values so that they don't need to be regenerated every time.

popematt:
**Issue #, if available:**

Fixes #215

**Description of changes:**

Listed by commit:

* Adds tests for Ion Schema 1.0 with transitive imports enabled
  * As I'm working on fixing imports, I don't want to cause any regressions for transitive imports
* Clean up IonSchemaVersion and remove incorrect function implementation
  * One of the functions was just wrong, so it's removed.
  * Some other functions marked internal or moved to test utilities
  * There's multiple definitions of the version marker regex around the codebase, so I consolidated them
* Adds getOrNull to SchemaCache
  * Allows us to check if a schema is already cached without having to put something in (as with `getOrPut`)
* Adds SchemaInternal interface
  * Removes/replaces `SchemaImpl`, which was a weird interface with a pseudo-contructor (the `operator fun invoke(...)`)
  * Moves the creation of Schema instances into `IonSchemaSystemImpl`
  * Refactors _everything_ that's not part of the public API to use `SchemaInternal` and `TypeInternal` instead of `Schema` and `Type` respectively. 
* Refactor Occurs and Fields constraints (fix #215)
  * Cleans up some unused code
  * Makes `Occurs` non-validating, and has a stateful, validating "copy" called `OccursValidator`.

N/A


ion-schema/src/main/kotlin/com/amazon/ionschema/SchemaCache.kt
```diff
@@ -19,10 +22,47 @@ interface SchemaCache {
...
+
     /**
      * Invalidates the cache entry at [key], if one exists.
      *
      * @param[key] the cache key to invalidate
      */
     fun invalidate(key: String)
 }
+
+private object UnimplementedSchema : Schema {
```
desaikd, line 53: Can we add a comment explaining the use case for this?

desaikd (approved):

Issue #209, opened 2022-09-29
Imported types cannot be resolved for cyclic imports


popematt:
Types cannot be resolved when there is a cycle of imports _and_ a type is actually used from each imported schema. The same behavior happens even if the types are imported by name or with an alias.
https://gist.github.com/popematt/6455c6391cdc47b14910caa7460c1ab4

When the cycle is made of inline imports, as in this example, a `StackOverflowError` occurs.
https://gist.github.com/popematt/098cf64c8a7a7f28cbf14253f2c41246

I tested this with commits fc5c91f (v1.2.0), cf795c9 (v1.2.1), and with the tip of `master` to see if there was a specific point in time when a regression occurred from #157, but it seems that #157 did not cover these particular cases.

PR #234, opened 2023-02-28
Adds DeferredReferenceManager and SchemaContentCache

popematt:
### Issue #, if available:

Partial progress for fixing #209 

#### General strategy:
* The current implementation does a depth first traversal, recursively loading schemas until it either hits a leaf node, or it discovers a cycle. It also eagerly loads all imported types, and so the types in Schema A cannot be resolved until all of its imports are loaded. When there's a cycle, we end up with a chicken and egg problem. Schema A cannot load its types until its imports from Schema B are loaded, and Schema B cannot load its types until its imports from Schema A are loaded.
* New implementation
  * A schema is loaded in two phases.
  * The first phase simply loads the Ion stream into a `SchemaContent` object. This is enough for us to inspect the names of the types declared by the schema, and as a result we can be sure that a type _exist_ in a schema before the schema is fully loaded.
  * The second phase goes from `SchemaContent` to `SchemaInternal`. This is where we ensure that the schema is syntactically and semantically valid, and we construct all of the `Type` objects.
  * When a schema is undergoing the second phase of loading, it only requires its dependencies to be in the first phase of loading. However, this means that we cannot resolve any imported types right away; it must be deferred until after the imported schema had undergone the second phase of loading.
  * This is where the `DeferredReferenceManager` comes into play. It is responsible for collecting deferred type references across multiple schemas, and ensuring that all of the references are resolved. DeferredReferences point to types that are guaranteed to exist, so they are only unresolvable if the target schema/type has some invalid syntax or has its own bad imports.
  * If any DeferredReferences cannot be resolved, then _all_ schemas that were loaded using that particular `DeferredReferenceManager` instance are evicted from the `IonSchemaSystem`'s `SchemaCache` so that the `SchemaCache` isn't polluted with unresolved/invalid schemas. The same schemas are also evicted from the `SchemaContentCache` so that if they get changed on disk, they can be reloaded properly.

The new implementation intentionally ignores the problem of cycles that are created using transitive imports. It is possible that this may work for some transitive import cycles, but that was a non-goal of this PR.

#### Description of changes in this PR:

 * Adds `DeferredReferenceManager` and `SchemaContentCache`
 * Also adds the data types that these classes manage
 * Also adds unit tests

### Related PRs in ion-schema, ion-schema-tests, ion-schema-schemas:

None.

----

### DIFF
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/DeferredReference.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/DeferredReference.kt
@@ -1,0 +1,130 @@
+/*
+ * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.ionschema.internal
+
+import com.amazon.ion.IonSymbol
+import com.amazon.ion.IonValue
+import com.amazon.ionschema.InvalidSchemaException
+import com.amazon.ionschema.Violations
+
+/**
+ * The classes in this file are an implementation detail of [DeferredReferenceManager], and should not be used outside
+ * [DeferredReferenceManager] except for testing. To properly encapsulate [DeferredReference] and its implementations,
+ * we would have to make them private inside [DeferredReferenceManager], but that makes it difficult to thoroughly test
+ * these classes. In order to directly test the private classes, we would have to use reflection to modify the class
+ * visibility in the unit tests. However, this will not work if the ClassLoader has a SecurityManager that prohibits
+ * changing visibility, and so the tests would be brittle.
+ *
+ * Instead, we have a [RequiresOptIn] annotation for all classes in this file. If you try to use [DeferredReference] or
+ * any implementations without explicitly opting in, it will result in a compile-time error. Opting in requires
+ *
+ * ```
+ * @OptIn(DeferredReferenceManagerImplementationDetails::class)
+ * ```
+ *
+ * Essentially, this is like a much stricter version of the @VisibleForTesting annotation that is commonly used in Java.
+ */
+@RequiresOptIn(
+    message = "Do not use DeferredReference and its implementations outside DeferredReferenceManager or unit tests.",
+    level = RequiresOptIn.Level.ERROR
+)
+@Retention(AnnotationRetention.BINARY)
+internal annotation class DeferredReferenceManagerImplementationDetails
+
+/**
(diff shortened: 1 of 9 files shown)


ion-schema/src/main/kotlin/com/amazon/ionschema/internal/SchemaContentCache.kt
```diff
@@ -0,0 +1,87 @@
...
+import com.amazon.ion.IonSymbol
+import com.amazon.ion.IonValue
+import com.amazon.ionschema.InvalidSchemaException
+
+/**
+ * A class that caches the uninterpreted content of a Schema document. This enables us to determine which types are
+ * declared in a schema prior to having a [SchemaInternal] instance for the schema, which is especially useful for
+ * checking the validity of a type reference without having to resolve the type reference to a [TypeInternal] instance.
+ */
+internal class SchemaContentCache constructor(
```
desaikd, line 27: (Question) Where is this `SchemaContentCache` used?  Is this used by `IonSchemaSystem` while resolving a schema?

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/DeferredReference.kt
```diff
@@ -0,0 +1,93 @@
...
+import com.amazon.ion.IonSymbol
+import com.amazon.ion.IonValue
+import com.amazon.ionschema.InvalidSchemaException
+import com.amazon.ionschema.Violations
+
+/**
+ * Represents a [TypeInternal] that is known to be declared in a schema document, but the [TypeInternal] instance has
+ * not necessarily been created yet. A [DeferredReference] can be resolved or unresolved. When it is unresolved,
+ * it is unknown whether the referred [TypeInternal] instance exists. When it is resolved, this [DeferredReference]
+ * is able to return and/or delegate validation to the referred [TypeInternal] instance.
```
desaikd, line 27: (Suggestion) Reading this it seems like we will always get a `TypeInternal`  but that might not be true in the cases where it is unsolvable. I think when it is resolved it has 2 possibilities either get `TypeInternal` or null.
Do you think we should add that in the comment here?

```suggestion
 * it is unknown whether the referred [TypeInternal] instance exists. When it is resolved, this [DeferredReference]
 * is able to return and/or delegate validation to the referred [TypeInternal] instance and if it not resolvable then it will return `null`. 
```

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/SchemaContentCache.kt
popematt: Yes, this will be used by the `IonSchemaSystem`, but it will be while loading the content of a schema. That part will come in a later PR.

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/DeferredReference.kt
popematt: What I am trying to say is that this class represents a `type::{ name: foo, ... }` that _is_ known to exist. However, the `TypeInternal` might not be instantiated yet. If it is not instantiated yet, then we cannot resolve this `DeferredReference`. Calling `resolve()` should only be done by the `DeferredReferenceManager` so that we can be sure it happens only at the correct time.

I will try to refactor these classes to better encapsulate the behavior, and I will fix the documentation.

desaikd: That would be great. Thanks.

desaikd (approved):

PR #238, opened 2023-03-03
Fixes cyclical import resolution for SchemaImpl_2_0 header imports

popematt:
**Issue #, if available:**

Continuation of fixing #209 

**Description of changes:**

For overall strategy, see #234 

* `SchemaImpl_2_0` gets a considerable rewrite of how imports are handled so that imported schemas can be lazily loaded. It uses the DeferredReferenceManager to create all imported type references. Also treats the core types as if they are imported (which they are) which cuts down on extra code branching.
* `ImportImpl` now resolves the schema lazily, so that we can load schemas in a loop rather than recursively. (This is what required the one-line change in `SchemaImpl_1_0`.)
* Adds `SchemaImpl_2_0_Test`
* Adds an `internal` function to get the core types for any supported ISL version to `IonSchemaSystemImpl`
* Updates `IonSchemaTestsRunner` so that it now runs the header import cycles tests for ISL 2.0.

None.

### DIFF
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/ImportImpl.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/ImportImpl.kt
@@ -21,14 +21,23 @@ import com.amazon.ionschema.Type
 
 /**
  * Implementation of [Import] for all user-provided ISL.
+ *
+ * This is no longer used to support any of the internal functionality of Ion Schema Kotlin. It exists only for the
+ * [Schema.getImports] and [Schema.getImport] functions.
+ *
+ * Do not call any functions of this class (directly or indirectly) from within the init block of _any_ [Schema]
+ * implementation. If you do, the call may fail, depending on the order that schema imports are resolved, which is not
+ * guaranteed to be stable.
+ *
+ * This class (and [Import]) should be removed in v2.0.0 in favor of a map of schemaId to [Type] (or similar).
  */
 internal class ImportImpl(
     override val id: String,
-    private val schema: Schema,
+    private val schemaProvider: () -> Schema,
     private val types: Map<String, Type>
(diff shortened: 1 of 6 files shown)


ion-schema/src/main/kotlin/com/amazon/ionschema/internal/ImportImpl.kt
```diff
@@ -21,14 +21,23 @@ import com.amazon.ionschema.Type
...
  * Implementation of [Import] for all user-provided ISL.
+ *
+ * This is no longer used to support any of the internal functionality of Ion Schema Kotlin. It exists only for the
+ * [Schema.getImports] and [Schema.getImport] functions.
+ *
+ * Do not call any functions of this class (directly or indirectly) from within the init block of _any_ [Schema]
+ * implementation. If you do, the call may fail, depending on the order that schema imports are resolved, which is not
+ * guaranteed to be stable.
+ *
+ * This class (and [Import]) should be removed in v2.0.0 in favor of a map of schemaId to [Type] (or similar).
```
tgregg, line 32: Should the prohibited methods throw? Or can they still be safely called by the user?

tgregg (approved):

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/ImportImpl.kt
popematt: They can still be called (safely) by the user. The methods will only throw if you call them during the `init { }` block of a `SchemaInternal` implementation.

popematt: And this particular class is hidden from the user. The user can only see the `Import` interface, so the user won't see this warning.

PR #240, opened 2023-03-04
Fixes cyclical import resolution for SchemaImpl_1_0 header imports

popematt:
*Issue #, if available:*

Continuation of #209 

*Description of changes:*

For overall strategy, see #234

This is the the equivalent to #238, but this is for `SchemaImpl_1_0`.

* Gets rid of schemaImportSet in IonSchemaSystemImpl in favor of just catching a StackOverflowException.
* Updates import logic in SchemaImpl_1_0 to use new DeferredReferences, and fallback to the old logic if transitive imports are enabled.
* Updates IonSchemaTestsRunner to skip the imports/cycle tests when running tests with transitive import enabled
* Updates `ion-schema-tests` submodule to the latest commit
  * Skips an unrelated failing test for ISL 2.0.

*Related PRs in ion-schema, ion-schema-tests, ion-schema-schemas:*

See changed files for changes in ion-schema-tests.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #241, opened 2023-03-06
Fixes cyclical import resolution for inline imports

### LINKED ISSUE #209: Imported types cannot be resolved for cyclic imports

popematt:
**Issue #, if available:**

Fixes #209 

**Description of changes:**

For overall strategy, see #234 

* `TypeReference.kt` is updated to use `DeferredReferenceManager` and `TypeReferenceDeferred` is removed. It's a slightly larger change than it maybe needed to be because I wanted to separate the inline import and the inline type definition cases into separate helper functions.
* `addDeferredType()` is removed from `SchemaInternal` and implementations as well as all other `TypeReferenceDeferred` code.
* Updates `IonSchemaTestsRunner` to no longer skip the inline imports cycle tests.

Cyclic import tests for ISl 1.0: https://github.com/amazon-ion/ion-schema-tests/pull/48
The tests for ISl 2.0 are essentially the same.
These changes were already incorporated by updating the git submodule in #240 .

### DIFF
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/SchemaCore.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/SchemaCore.kt
@@ -81,7 +81,6 @@ internal class SchemaCore(
     override fun newType(isl: String) = throw UnsupportedOperationException()
     override fun newType(isl: IonStruct) = throw UnsupportedOperationException()
     override fun plusType(type: Type) = throw UnsupportedOperationException()
-    override fun addDeferredType(typeRef: TypeReferenceDeferred) = throw UnsupportedOperationException()
 }
 
 private const val CORE_TYPES =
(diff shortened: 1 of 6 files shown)


tgregg (approved):
Nice

PR #247, opened 2023-03-16
Fix multiple bugs related to nullability annotations

### LINKED ISSUE #246: `nullable::` annotation does not work for imported types
The `nullable::` annotation does not work on imported types. (It results in an `IllegalStateException` being thrown when trying to load the schema.) This is a regression that was introduced in the 1.5.0 release.

popematt:
**Issue #, if available:**

Fixes #246 

**Description of changes:**

Basically, when constructing a `TypeNullable`, in the `init` block, it tries to check the Ion type of the type that it is decorating. If it determines that the Ion type is `IonType.DATAGRAM`, it throws an exception because there is no such thing as a null datagram. There is no nice way to fix this, so this errs on the side of being overly permissive—if we run into an unresolved type reference, the `TypeNullable` initializer swallows the exception and just moves on.

While I was working on this, I also discovered that `ion-schema-kotlin` is incorrectly allowing `nullable::` and `$null_or::` annotations on variably occurring type definitions (i.e. types with `occurs` in them), so I fixed that as well as refactoring `Occurs` while I was in there.

See 🗺️ comments for details about specific code changes.

https://github.com/amazon-ion/ion-schema-tests/pull/49

### DIFF
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/DeferredReference.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/DeferredReference.kt
@@ -73,12 +73,14 @@ internal interface DeferredReference : TypeInternal {
 
     companion object {
         private fun illegalMethodCallBeforeResolving(): Nothing {
-            throw IllegalStateException("This should be unreachable because this method should not be called before the call to DeferredReferenceManager.resolve() is complete.")
+            throw UnresolvedTypeReferenceException("This should be unreachable because this method should not be called before the call to DeferredReferenceManager.resolve() is complete.")
         }
     }
     override fun getBaseType(): TypeBuiltin = if (isResolved()) resolve().getBaseType() else illegalMethodCallBeforeResolving()
     override fun isValidForBaseType(value: IonValue): Boolean = if (isResolved()) resolve().isValidForBaseType(value) else illegalMethodCallBeforeResolving()
     override fun validate(value: IonValue, issues: Violations) = if (isResolved()) resolve().validate(value, issues) else illegalMethodCallBeforeResolving()
+
+    class UnresolvedTypeReferenceException(message: String) : IllegalStateException(message)
 }
 
 /**
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/SchemaContentCache.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/SchemaContentCache.kt
@@ -18,6 +18,7 @@ package com.amazon.ionschema.internal
 import com.amazon.ion.IonSymbol
 import com.amazon.ion.IonValue
 import com.amazon.ionschema.InvalidSchemaException
+import com.amazon.ionschema.internal.util.withoutTypeAnnotations
 
 /**
  * A class that caches the uninterpreted content of a Schema document. This enables us to determine which types are
@@ -75,7 +76,7 @@ internal class SchemaContentCache constructor(
      * into this instance of [SchemaContentCache].
      */
     fun doesSchemaDeclareType(schemaId: String, typeId: IonSymbol): Boolean {
-        return runCatching { typeId in getSchemaContent(schemaId).declaredTypes }.getOrElse { false }
+        return runCatching { typeId.withoutTypeAnnotations() in getSchemaContent(schemaId).declaredTypes }.getOrElse { false }
     }
 
     /**
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/TypeReference.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/TypeReference.kt
@@ -61,7 +61,7 @@ internal class TypeReference private constructor() {
                 islRequire(illegalAnnotations.isEmpty()) { "Illegal annotations $illegalAnnotations on type reference: $ion" }
             }
 
-            return when (ion) {
+            val theType = when (ion) {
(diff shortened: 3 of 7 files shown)


ion-schema/src/main/kotlin/com/amazon/ionschema/internal/DeferredReference.kt
```diff
@@ -73,12 +73,14 @@ internal interface DeferredReference : TypeInternal {
...
         private fun illegalMethodCallBeforeResolving(): Nothing {
-            throw IllegalStateException("This should be unreachable because this method should not be called before the call to DeferredReferenceManager.resolve() is complete.")
+            throw UnresolvedTypeReferenceException("This should be unreachable because this method should not be called before the call to DeferredReferenceManager.resolve() is complete.")
         }
     }
     override fun getBaseType(): TypeBuiltin = if (isResolved()) resolve().getBaseType() else illegalMethodCallBeforeResolving()
     override fun isValidForBaseType(value: IonValue): Boolean = if (isResolved()) resolve().isValidForBaseType(value) else illegalMethodCallBeforeResolving()
     override fun validate(value: IonValue, issues: Violations) = if (isResolved()) resolve().validate(value, issues) else illegalMethodCallBeforeResolving()
+
+    class UnresolvedTypeReferenceException(message: String) : IllegalStateException(message)
```
popematt, line 83: 🗺️ New exception type is so that we can safely catch and swallow it elsewhere without accidentally catching and swallowing exceptions for unrelated problems.

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/SchemaContentCache.kt
```diff
@@ -75,7 +76,7 @@ internal class SchemaContentCache constructor(
      * into this instance of [SchemaContentCache].
      */
     fun doesSchemaDeclareType(schemaId: String, typeId: IonSymbol): Boolean {
-        return runCatching { typeId in getSchemaContent(schemaId).declaredTypes }.getOrElse { false }
+        return runCatching { typeId.withoutTypeAnnotations() in getSchemaContent(schemaId).declaredTypes }.getOrElse { false }
```
popematt, line 79: 🗺️ Comparing the `typeId` symbol with the annotations intact was causing e.g. `nullable::foo` to not be found even though `foo` type existed.

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/TypeReference.kt
```diff
@@ -80,22 +84,23 @@ internal class TypeReference private constructor() {
                 is IonSymbol -> handleSymbol(ion, schema, referenceManager)
                 else -> throw InvalidSchemaException("Unable to resolve type reference '$ion'")
             }
+
+            val maybeNullableType = handleNullable(ion, schema, theType)
```
popematt, line 88: 🗺️ There's a lot of branching in this class to handle different sorts of type references. I refactored the class so that `handleNullable()` only needs to be called in one place (so that it doesn't get missed for any branches).

Basically, this means that all other calls to `handleNullable()` are removed, and all of the private functions now return `TypeInternal` instead of `() -> TypeInternal`.

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/constraint/Fields.kt
```diff
@@ -65,7 +64,7 @@ internal class Fields(
         // Forces the field definitions to be validated
         fieldConstraints = ionStruct.associateBy(
             { it.fieldName },
-            { Occurs(it, schema, referenceManager, OPTIONAL, isField = true) }
+            { Occurs(it, schema, referenceManager, isField = true) }
```
popematt, line 67: 🗺️ Just removing the now-unnecessary constructor arg for `Occurs`.

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/constraint/Occurs.kt
```diff
@@ -92,36 +83,12 @@ internal class Occurs(
...
-                } else {
-                    defaultRange
-                }
-            } else {
-                defaultRange
-            }
+        val occurs: IonValue? = (ion as? IonStruct)?.takeIf { !ion.isNullValue }?.get("occurs")
 
-        val tmpIon = if (ion is IonStruct && occurs != null) {
-            ion.cloneAndRemove("occurs")
```
popematt, line 109: 🗺️ Removing this line is the only change in this file that is part of the actual fix for prohibiting `nullable` and `$null_or` on variably occurring type definitions. All of the rest of the changes in this file are just cleaning up some overly complicated code.

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/util/IonValueExtensions.kt
```diff
@@ -24,9 +24,9 @@ import com.amazon.ion.IonValue
  * If this value has no annotations, returns `this`;
  * otherwise, returns a clone of `this` with the annotations removed.
  */
-internal fun IonValue.withoutTypeAnnotations() =
+internal fun <T : IonValue> T.withoutTypeAnnotations(): T =
```
popematt, line 27: 🗺️ Receiver is made generic so that when you call this function on e.g. an IonSymbol, the result is still an IonSymbol (as opposed to an IonValue that needs to be cast back to IonSymbol).

tgregg (approved):

Issue #250, opened 2023-04-17
Add a POJO (or similar) model for Ion Schema


popematt:
Having a properly typed model will make it nicer to manipulate the schema definitions for things such as:
* converting to/from other schema formats
* comparing two schemas
* writing a linter tool for schemas

Issue #109, opened 2019-04-24
refactor range implementations in software.amazon.ionschema.internal.util


pbcornell:
In particular, note that:
* IntRange duplicates functionality provided by RangeInt, but offers some nice simplifications, as well as exposes the right interface for use by StateMachine
* RangeFactory.rangeOf() is awkward;  perhaps the codebase is better off without a factory here
* the heavy use of delegation tends to be confusing;  is there a way to simplify (or at least clarify) how this works?

PR #251, opened 2023-04-17
Adds ISL model

popematt:
*Issue #, if available:*

#250 

Also, partial work on #109 in that this PR has new range implementations that will eventually replace the existing range implementations.

*Description of changes:*

Adds a model for Ion Schema. Currently the model is opt-in only—gated by the `@ExperimentalIonSchemaModel` annotation—and subject to breaking changes.

A few questions/comments for reviewers:
* I have decided to use the term "Type Argument" instead of "type reference" to refer to the thing that you provide to a constraint (such as `element`, `all_of`, etc.). This is a departure from the spec, but I believe it will eliminate the confusion between a "reference" and a so-called "type reference" (that could be a "reference", an inline import, or an inline type). I plan to update the spec to use the new terminology.
* A large chunk of code from the original Regex constraint impl is moved to a separate `regex.kt` utility file so that the regex validation code can be shared (for now) between the model and the existing constraint impl.
* Marking classes as `data` classes means that we can't hide implementation details as well. `data` classes always indirectly expose the primary constructor through the auto-generated `copy()` implementation, and `data` classes always have destructuring `componentN()` functions. I would like to eventually make all of the constructors `internal` and provide a version-safe builder for constructing the ISL model. The benefit of `data` is that we also get autogenerated `equals()` and `hashCode()` functions. Should we _not_ use data classes so as not to leak implementation through the `copy`, etc. functions? 
* Right now the constraint classes are nested in the `Constraint` interface. Thoughts? Should they be moved to a `...ionschema.model.constraints` packages?
* I plan to have an `internal sealed interface SealedConstraint` (once we upgrade the language level). Is it worth making `SealedConstraints` public? If it is public, it would have to be gated by an opt-in annotation because every time we add a new constraint, it would be a breaking change. (See "TODO" comment in `Constraint.kt` for more discussion.)

N/A

### DIFF
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/model/Constraint.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/model/Constraint.kt
@@ -1,0 +1,235 @@
+package com.amazon.ionschema.model
+
+import com.amazon.ion.IonValue
+import com.amazon.ionschema.IonSchemaVersion
+import com.amazon.ionschema.internal.util.validateRegexPattern
+import kotlin.text.Regex as KRegex
+
+/**
+ * Marker interface for all Constraint implementations.
+ *
+ * This implementation of Ion Schema does not support custom constraints. Do not implement this interface.
+ */
+@ExperimentalIonSchemaModel
+interface Constraint {
+
+    /**
+     * Represents the `all_of` constraint.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#all_of) and
+     * [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#all_of).
+     */
+    data class AllOf(val types: TypeArgumentList) : Constraint
+
+    /**
+     * Represents the `annotations` constraint for Ion Schema 1.0.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#annotations).
+     */
+    data class AnnotationsV1(val annotations: List<Annotation>, val closed: Boolean, val ordered: Boolean) :
+        Constraint {
+        class Annotation(val text: String, val required: Boolean)
+    }
+
+    /**
+     * Represents the `annotations` constraint from Ion Schema 2.0 onwards.
+     * See relevant section in [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#annotations).
+     */
+    data class AnnotationsV2(val type: TypeArgument) : Constraint
+
+    /**
+     * Represents the `any_of` constraint.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#any_of) and
+     * [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#any_of).
+     */
+    data class AnyOf(val types: TypeArgumentList) : Constraint
+
+    /**
+     * Represents the `byte_length` constraint.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#byte_length) and
+     * [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#byte_length).
+     */
+    data class ByteLength(val range: DiscreteIntRange) : Constraint {
+        init {
+            range.start?.let { require(it >= 0) }
+        }
+    }
+
+    /**
+     * Represents the `codepoint_length` constraint.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#codepoint_length) and
+     * [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#codepoint_length).
+     */
+    data class CodepointLength(val range: DiscreteIntRange) : Constraint {
+        init {
+            range.start?.let { require(it >= 0) }
+        }
+    }
+
+    /**
+     * Represents the `container_length` constraint.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#container_length) and
+     * [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#container_length).
+     */
+    data class ContainerLength(val range: DiscreteIntRange) : Constraint {
+        init {
+            range.start?.let { require(it >= 0) }
+        }
+    }
+
+    /**
+     * Represents the `contains` constraint.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#contains) and
+     * [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#contains).
+     */
+    data class Contains(val values: List<IonValue>) : Constraint
+
+    /**
+     * Represents the `element` constraint.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#element) and
+     * [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#element).
+     *
+     * The [distinct] modifier is only supported for Ion Schema 2.0 and higher.
+     */
+    data class Element(val type: TypeArgument, val distinct: Boolean = false) : Constraint
+
+    /**
+     * Represents the `exponent` constraint, introduced in Ion Schema 2.0.
+     * See relevant section in [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#exponent).
+     */
+    data class Exponent(val range: DiscreteIntRange) : Constraint
+
+    /**
+     * Represents the `field_names` constraint, introduced in Ion Schema 2.0.
+     * See relevant section in [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#field_names).
+     */
+    data class FieldNames(val type: TypeArgument, val distinct: Boolean = false) : Constraint
+
+    /**
+     * Represents the `fields` constraint.
+     * Beware when reading from ISL—the ISL representation of [closed] was changed between ISL 1.0 and ISL 2.0.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#fields) and
+     * [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#fields).
+     */
+    data class Fields(val fields: Map<String, VariablyOccurringTypeArgument>, val closed: Boolean) : Constraint
+
+    /**
+     * Represents the `ieee754_float` constraint, introduced in Ion Schema 2.0.
+     * See relevant section in [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#ieee754_float).
+     */
+    data class Ieee754Float(val format: Ieee754InterchangeFormat) : Constraint
+
+    /**
+     * Represents the `not` constraint.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#not) and
+     * [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#not).
+     */
+    data class Not(val type: TypeArgument) : Constraint
+
+    /**
+     * Represents the `one_of` constraint.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#one_of) and
+     * [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#one_of).
+     */
+    data class OneOf(val types: TypeArgumentList) : Constraint
+
+    /**
+     * Represents the `ordered_elements` constraint.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#ordered_elements) and
+     * [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#ordered_elements).
+     */
+    data class OrderedElements(val types: List<VariablyOccurringTypeArgument>) : Constraint
+
+    /**
+     * Represents the `precision` constraint.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#precision) and
+     * [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#precision).
+     */
+    data class Precision(val range: DiscreteIntRange) : Constraint {
+        init {
+            range.start?.let { require(it >= 1) }
+        }
+    }
+
+    /**
+     * Represents the `regex` constraint.
+     *
+     * Ion Schema regular expressions are a subset of ECMA-262.
+     * The allowed subset of ECMA-262 regular expressions was changed between ISL 1.0 and ISL 2.0.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#regex) and
+     * [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#regex).
+     *
+     * See also [_Comparison of regular expression engines_](https://en.wikipedia.org/wiki/Comparison_of_regular_expression_engines#Language_features)
+     * for the high-level differences between ECMA (JavaScript) and Java regular expressions.
+     */
+    data class Regex(
+        val pattern: String,
+        val caseInsensitive: Boolean = false,
+        val multiline: Boolean = false,
+        val ionSchemaVersion: IonSchemaVersion = IonSchemaVersion.v2_0,
+    ) : Constraint {
+        internal val compiled = run {
+            require(pattern.isNotEmpty())
+            val opts = mutableSetOf<RegexOption>()
+            if (caseInsensitive) opts.add(RegexOption.IGNORE_CASE)
+            if (multiline) opts.add(RegexOption.MULTILINE)
+            KRegex(validateRegexPattern(pattern, ionSchemaVersion), opts)
+        }
+    }
+
+    /**
+     * Represents the `scale` constraint.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#scale).
+     *
+     * This constraint was replaced by `exponent` in Ion Schema 2.0.
+     * To convert to [Exponent], used [Scale.toExponentConstraint].
+     */
+    data class Scale(val range: DiscreteIntRange) : Constraint {
+        fun toExponentConstraint() {
+            Exponent(range.negate())
+        }
+    }
+
+    /**
+     * Represents the `timestamp_offset` constraint.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#timestamp_offset) and
+     * [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#timestamp_offset).
+     *
+     * @see TimestampOffsetValue
+     */
+    data class TimestampOffset(val offsets: List<TimestampOffsetValue>) : Constraint
+
+    /**
+     * Represents the `timestamp_precision` constraint.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#timestamp_precision) and
+     * [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#timestamp_precision).
+     *
+     * @see [TimestampPrecisionRange]
+     */
+    data class TimestampPrecision(val range: TimestampPrecisionRange) : Constraint
+
+    /**
+     * Represents the `type` constraint.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#type) and
+     * [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#type).
+     */
+    data class Type(val type: TypeArgument) : Constraint
+
+    /**
+     * Represents the `utf8_byte_length` constraint.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#utf8_byte_length) and
+     * [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#utf8_byte_length).
+     */
+    data class Utf8ByteLength(val range: DiscreteIntRange) : Constraint {
+        init {
+            range.start?.let { require(it >= 0) }
+        }
+    }
+
+    /**
+     * Represents the `valid_values` constraint.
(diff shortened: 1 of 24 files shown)


ion-schema/src/main/kotlin/com/amazon/ionschema/model/util.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Negates the boundaries of the range, and swaps their position to keep the lower number in the `start` position.
+ *
+ * For example:
+ * ```
+ * val range1 = DiscreteIntRange(-2, 10)
+ * val range2 = DiscreteIntRange(-10, 2)
+ * assertEquals(range1.negate(), range2)
+ * ```
+ */
+fun DiscreteIntRange.negate(): DiscreteIntRange {
```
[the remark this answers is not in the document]
popematt: This is used in the `Scale.toExponent()` function. (I guess it's not necessary to add that function in yet, but I wanted to get all of the range stuff done with.)

ion-schema/src/main/kotlin/com/amazon/ionschema/model/TimestampPrecisionValue.kt
```diff
@@ -0,0 +1,41 @@
+package com.amazon.ionschema.model
+
+/**
+ * Represents the available arguments for the `timestamp_precision` constraint.
+ * @see Constraint.TimestampPrecision
+ */
+enum class TimestampPrecisionValue(internal val id: Int) {
```
rmarrowstone, line 7: Kotlin internal doesn't provide protection to consumers writing java (at least it didn't in previous versions). I would say just make it private or not (the nice part about kotlin is we can always set an `id` property later if need be).

ion-schema/src/main/kotlin/com/amazon/ionschema/model/TypeArgument.kt
```diff
@@ -0,0 +1,46 @@
+package com.amazon.ionschema.model
+
+/**
+ * A TypeArgument represents (defines or references) a Type, allowing the Type to be used as an argument for a constraint.
+ *
+ * **_Do not implement this interface_**—It will become a sealed interface once `ion-schema-kotlin` is updated to use
+ * language version 1.5.
+ */
+@ExperimentalIonSchemaModel
+sealed class TypeArgument {
```
rmarrowstone, line 10: I like `Argument` or `Parameter` maybe for this.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/TypeArgument.kt
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * Nullability modifiers for [TypeArgument].
+     */
+    enum class Nullability {
+        /**
+         * No special treatment of null values.
+         */
+        None,
+        /**
+         * Ion Schema 1.0 nullability. See https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#core-types
```
rmarrowstone, line 22: I'm pretty inclined to make modeling ISL 1 a non-goal.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/UserReservedFields.kt
```diff
@@ -0,0 +1,11 @@
+package com.amazon.ionschema.model
+
+/**
+ * The collection of field names that are reserved by the user as open content fields.
+ * See relevant section in [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#open-content).
+ */
+data class UserReservedFields constructor(
+    val type: List<String> = emptyList(),
+    val header: List<String> = emptyList(),
+    val footer: List<String> = emptyList(),
```
rmarrowstone, line 10: hrrm. The problem is that Kotlin's List is read-only but not technically immutable. IIRC, like `internal` this is something that can be circumvented with java code. Not sure how protective we want to get, just making sure you're aware of it.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/ValidValue.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import com.amazon.ion.IonValue
+
+/**
+ * Represents an argument to the `valid_values` constraint.
+ * @see [Constraint.ValidValues]
+ */
+sealed class ValidValue {
+    /**
+     * A single Ion value. May not be annotated.
```
rmarrowstone, line 11: Do we need a note about the value the model holds and it is an Ion equivalence comparison?

rmarrowstone: Maybe just like a warning to someone who is introspecting the model so they are clear on the semantics?

ion-schema/src/test/kotlin/com/amazon/ionschema/model/ContinuousRangeTest.kt
```diff
@@ -0,0 +1,168 @@
...
+    }
+    @ParameterizedTest(name = "[constructor] ContinuousRange({0}, {1}) should not be valid")
+    @MethodSource("invalidRanges")
+    fun `ContinuousRange should be invalid`(min: ContinuousRange.Limit<Double>?, max: ContinuousRange.Limit<Double>?) {
+        assertThrows<IllegalArgumentException> { ContinuousRange(min, max) }
+    }
+
+    @CsvSource(
+        // lower limit | high limit | value | is contained?
+        "CLOSED,  2.0, CLOSED,  4.0,   1.0, false",
```
rmarrowstone, line 26: qq: why can we not just give kotlin lists for this? is it clearer as csv?

rmarrowstone (commented):
Good stuff! I don't see any major blockers, would like to understand if we can descope modeling the ISL 1 stuff.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/Constraint.kt
```diff
@@ -0,0 +1,236 @@
+package com.amazon.ionschema.model
+
+import com.amazon.ion.IonValue
+import com.amazon.ionschema.IonSchemaVersion
+import com.amazon.ionschema.internal.util.validateRegexPattern
+import kotlin.text.Regex as KRegex
+
+/**
+ * This internal only interface allows us to use an exhaustive `when` within this library.
+ * It will become `sealed` when we update to kotlin language version = 1.5 or higher.
```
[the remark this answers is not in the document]
popematt: Mostly just because we haven't determined whether an upgrade would require a new major version, and we haven't had a particularly strong impetus before now.

rmarrowstone (commented):
My only comment on the `data` class question is: it looks like (with few exceptions) all of the fields in the model are _public_ `val`s. I don't really see what there is to hide! If we need to add indirection later we can always redefine them as a property.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/Constraint.kt
```diff
@@ -0,0 +1,236 @@
...
+ *
+ * This is not visible to users of the library. Users of the library must use [Constraint], which is not sealed so that
+ * we can add new constraints (in new versions of Ion Schema) to this library in a way that is backwards compatible for
+ * consumers.
+ *
+ * TODO—some library users might prefer for their builds to break when they are not handling a new constraint. Should
+ * we make this public with the caveat that it requires opt in, and is subject to breaking changes?
```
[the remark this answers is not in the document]
popematt: I think we do not want to allow people to implement their own (for now). There was discussion about this as part of the ISL 2.0 spec work. We declined to introduce custom constraints (for now) because there was no concrete use case with which to justify the cost in complexity.

However, one thing we did decide is that _if_ we do add user-defined constraints, we need to have support for it in the Ion Schema Specification, or else we'll start running into problems with schemas not being portable (i.e. different behavior for the same inputs on differently configured systems). For example, a custom constraint using the "is_prime" field name—if Ion Schema Kotlin on one system has that constraint installed, and on another system, Ion Schema Kotlin does not have that constraint installed, then the same schema and the same data could have different validation results on two different machines.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/DiscreteRange.kt
```diff
@@ -0,0 +1,41 @@
...
+ * A range over a type that is a finite or countably infinite set. The values contained in the range include both of the
+ * limits—[start] and [endInclusive].
+ *
+ * A `DiscreteRange` is a _bounded, closed interval_ when both limits are non-null, and it is a _half-bounded interval_
+ * when one of the limits is `null`. At least one of [start] and [endInclusive] must be non-null.
+ *
+ * A `DiscreteRange` is allowed to be a _degenerate interval_ (i.e. `start == endInclusive`), but it may not be an
+ * empty interval (i.e. `start > endInclusive`).
+ */
+data class DiscreteRange<T : Comparable<T>>(val start: T?, val endInclusive: T?) {
```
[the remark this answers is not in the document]
popematt: Discrete Ranges don't require us to store whether the boundary value is exclusive because we can "translate" exclusive to +/-1 of the boundary value.

More importantly, however, is the fact that the logic for determining emptiness is different. With a discrete range, it is true that `∀n∈Z, ¬∃x∈Z, n<x & x<n+1`, but for a range over continuous values, it would not be empty. If `n` and `x` are both in the set of all rational numbers rather than the set of all integers, then we can trivially prove that there does exist at least one `x`, such as `x = n + 0.5`.

Finally, my admittedly subjective opinion is that it helps the reader/user understand how the ranges are to be used by keeping them distinct.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/Constraint.kt
```diff
@@ -0,0 +1,236 @@
...
+     * [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#all_of).
+     */
+    data class AllOf internal constructor(val types: TypeArgumentList) : SealedConstraint
+
+    /**
+     * Represents the `annotations` constraint for Ion Schema 1.0.
+     * See relevant section in [ISL 1.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-1-0/spec#annotations).
+     */
+    @Deprecated("The annotations constraint in Ion Schema 1.0 has poorly defined behavior.")
+    data class AnnotationsV1 internal constructor(val annotations: List<Annotation>, val closed: Boolean, val ordered: Boolean) : SealedConstraint {
```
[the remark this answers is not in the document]
popematt: All Ion Schema 2.0 implementations are required to be able to import/operate on ISL 1.0. I probably should not actually mark it as `@Deprecated`.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/SchemaDocument.kt
```diff
@@ -0,0 +1,40 @@
...
+ */
+@ExperimentalIonSchemaModel
+data class SchemaDocument internal constructor(
+    val id: String?,
+    val items: List<Item>
+) {
+    val ionSchemaVersion: IonSchemaVersion = items.firstOrNull { it !is Item.OpenContent }
+        ?.let { it as? Item.VersionMarker }
+        ?.value
+        ?: IonSchemaVersion.v1_0
```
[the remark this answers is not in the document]
popematt: I don't think there is any invariant that requires validation here.

From the [ISL Versioning Spec](https://amazon-ion.github.io/ion-schema/docs/isl-versioning#ion-schema-version-markers)
> The first value in the schema SHOULD be an ISL Version Marker; if a top-level ISL value is encountered before encountering an ISL version marker, that schema SHALL use ISL 1.0.

So, we're finding the first this that is not open content (i.e. is is a top-level ISL value). If that top-level value is a version marker, then we use that version. Otherwise, it is ISL 1.0.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/TypeArgument.kt
popematt: Some constraints have arguments that are not types, however, so I wanted to qualify it with the word "Type". Generally, "parameter" describes a function's signature whereas "argument" refers to the variables/values being passed to a function, and that's how I arrived at `TypeArgument`

ion-schema/src/main/kotlin/com/amazon/ionschema/model/TimestampPrecisionValue.kt
popematt: I thought the Kotlin compiler did some name mangling for `internal` items (protection through obscurity), and if so, I would be satisfied with that. However, I can refactor this so that `id` can be private.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/ValidValue.kt
popematt: > Do we need a note about the value the model holds and it is an Ion equivalence comparison?

Sure, I can add that. It is using Ion equivalence.

ion-schema/src/test/kotlin/com/amazon/ionschema/model/ContinuousRangeTest.kt
popematt: If we use Kotlin lists, then I have to put the arguments in a `static` method in the `companion object`. Having them in CSV keeps them right by the test that actually uses the arguments. In some of the other tests, I've used the `@MethodSource` because the arguments were re-used across multiple tests or because the arguments were too complex or too many to hand-write as CSV.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/TypeArgument.kt
rmarrowstone: Yea. I should've been more clear that I agree with `Type` as the prefix.

popematt:
> My only comment on the `data` class question is: it looks like (with few exceptions) all of the fields in the model are _public_ `val`s. I don't really see what there is to hide! If we need to add indirection later we can always redefine them as a property.

I agree. I don't think there is anything to hide.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/UserReservedFields.kt
popematt: I am aware. If we're concerned about this, I could bring in the `kotlinx.collections.immutable` library.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/TypeArgument.kt
popematt: I don't think it's an option because of the requirement for ISL 2.0 implementations to also support ISL 1.0

ion-schema/src/main/kotlin/com/amazon/ionschema/model/TypeArgument.kt
```diff
@@ -0,0 +1,46 @@
+package com.amazon.ionschema.model
+
+/**
+ * A TypeArgument represents (defines or references) a Type, allowing the Type to be used as an argument for a constraint.
```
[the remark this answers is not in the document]
popematt: I will, once the link exists.

ion-schema/src/test/kotlin/com/amazon/ionschema/model/ContinuousRangeTest.kt
rmarrowstone: I like keeping them by the tests that uses the arguments. It wasn't (and still isn't) clear to me why that necessitates we use CsvSource except that apparently has some special mojo for this.

rmarrowstone: I guess it's just that there is no other readily provided annotation for providing data in this way. odd.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/DiscreteRange.kt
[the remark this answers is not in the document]
rmarrowstone: But whether a range is continuous or discrete depends on the type of arguments, right? I don't see any constraints that model that... so it looks like I can define a Discrete range of a Continuous (but Comparable) type and vice versa.

Further I don't see any definition in the spec about emptiness or degenerate-ness so I don't see where the requirement for the validation (and the need to split it) even comes from.

What I do see is what _appears to be_ duplicated code. 🤷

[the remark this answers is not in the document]
popematt: The test suite requires that ranges (at least for countable things like number of codepoints) are non-empty, and I believe that requirement has been part of the reference implementation since the initial release. https://github.com/popematt/ion-schema-tests/blob/master/ion_schema_2_0/constraints/codepoint_length.isl#L79

At this point, the non-empty range requirement is _de facto_ part of the specification, and the specification document should be updated to reflect that.

Degenerate ranges are fine. I could remove that bit of the sentence if it's too confusing to have it, though.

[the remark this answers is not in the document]
popematt: I have found a clean way to avoid the duplicated code, though, so I will remove the `DiscreteRange<T>` implementation.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/aliases.kt
```diff
@@ -0,0 +1,42 @@
...
+import com.amazon.ion.IonNumber
+import com.amazon.ion.IonValue
+import com.amazon.ion.Timestamp
+import java.math.BigDecimal
+
+/**
+ * Convenience alias for open content in schema headers, footers, and type definitions.
+ * Called 'OtherFields' to make it clear that this is for struct fields.
+ */
+typealias OtherFields = List<Pair<String, IonValue>>
```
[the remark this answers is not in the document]
popematt: This is independent of the schema version. It's just a container to hold extra struct fields.

ion-schema/src/test/kotlin/com/amazon/ionschema/model/SchemaDocumentTest.kt
```diff
@@ -0,0 +1,147 @@
...
+        val schema = SchemaDocument("schema.isl", listOf())
+        assertTrue(schema.declaredTypes.isEmpty())
+    }
+
+    @Test
+    fun `test get declaredTypes when some types present`() {
+        val type0 = NamedTypeDefinition(
+            "type0",
+            TypeDefinition(
+                constraints = listOf(),
```
desaikd, line 57: (Suggestion) Can we please add some tests for types that have constraints defined for ISL 1.0 and ISL 2.0.?

ion-schema/src/main/kotlin/com/amazon/ionschema/model/Constraint.kt
```diff
@@ -0,0 +1,235 @@
...
+     * Represents the `exponent` constraint, introduced in Ion Schema 2.0.
+     * See relevant section in [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#exponent).
+     */
+    data class Exponent(val range: DiscreteIntRange) : Constraint
+
+    /**
+     * Represents the `field_names` constraint, introduced in Ion Schema 2.0.
+     * See relevant section in [ISL 2.0 spec](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#field_names).
+     */
+    data class FieldNames(val type: TypeArgument, val distinct: Boolean = false) : Constraint
```
desaikd, line 104: (More information) Since `Constraint` has data classes for both ISL1.0 and ISL 2.0 constraints construction, how do we ensure that ISL2.0 constraints are not used to construct an ISL 1.0 type/schema?

ion-schema/src/test/kotlin/com/amazon/ionschema/model/SchemaDocumentTest.kt
popematt: My plan is to have builders that only allow you to create valid 1.0 or 2.0 types, so it will not be possible to create something with incompatible constraints. When I have those, it will make sense to test them in the way you are suggesting.

desaikd: I couldn't find any type tests with constraints in them. I was suggesting to have some tests for types with ISL 1.0 constraints and other tests for types with  ISL 2.0 constraints.

popematt: I see. There are no tests for types in the data model because `TypeDefinition` doesn't really have functionality to test. It's just plain data.

desaikd: Oh, but you are still asserting two type definitions below in this test. I think there should be some tests for checking how the data classes defined in `Constraint` work whether its separate tests for constraints  or as part of tests for types with constraints. If you have planned for this in a follow up PR then I am okay with what we have now but it might be worth mentioning this in an issue/TODO.

popematt: > Oh, but you are still asserting two type definitions below in this test. I think there should be some tests for checking how the data classes defined in Constraint work whether its separate tests for constraints or as part of tests for types with constraints.

Okay, so any tests regarding constraints would be done separately from this. This test class is unit tests for `SchemaDocument`, so the scope is limited to things that interact directly with `SchemaDocument`.

I am creating some empty types so that I have something to put in the schema, and I am asserting that they are the _same object_ (referential equality rather than value-based equality). Instead of using empty/dummy types, I can use mocks if that makes it more clear.

> If you have planned for this in a follow up PR then I am okay with what we have now but it might be worth mentioning this in an issue/TODO.

When there is functionality to be tested, I will either write the tests or I will create an issue/todo. Right now, there is nothing to be unit tested for `TypeDefinition`, so I don't even know what I would put into the issue/todo.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/Constraint.kt
desaikd: It seems like you have a plan to provide builders that ensure this so I resolving this.

ion-schema/src/test/kotlin/com/amazon/ionschema/model/SchemaDocumentTest.kt
desaikd: Thanks for explaining. It seems like these tests are just to ensure `SchemaDocument` is constructed properly.

desaikd (approved):

ion-schema/src/main/kotlin/com/amazon/ionschema/model/ContinuousRange.kt
```diff
@@ -0,0 +1,95 @@
...
+ */
+data class ContinuousRange<T : Comparable<T>>(val start: Limit<T>, val end: Limit<T>) {
+
+    sealed class Limit<T : Comparable<T>> {
+        abstract val value: T?
+
+        interface Bounded<T> { val value: T }
+        data class Closed<T : Comparable<T>>(override val value: T) : Limit<T>(), Bounded<T>
+        data class Open<T : Comparable<T>>(override val value: T) : Limit<T>(), Bounded<T>
+        class Unbounded<T : Comparable<T>> : Limit<T>() {
```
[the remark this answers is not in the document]
popematt: Actually, it doesn't play well with generics. `object Unbounded: Limit<Nothing>()` was giving me trouble getting the variances right, but you are right that objects are not as nice in Java.

ion-schema/src/main/kotlin/com/amazon/ionschema/model/DiscreteRange.kt
```diff
@@ -0,0 +1,50 @@
...
+ * A range over a type that is a finite or countably infinite set. The values contained in the range include both of the
+ * limits—[start] and [endInclusive].
+ *
+ * A `DiscreteRange` is a _bounded, closed interval_ when both limits are non-null, and it is a _half-bounded interval_
+ * when one of the limits is `null`. At least one of [start] and [endInclusive] must be non-null.
+ *
+ * A `DiscreteRange` is allowed to be a _degenerate interval_ (i.e. `start == endInclusive`), but it may not be an
+ * empty interval (i.e. `start > endInclusive`).
+ */
+class DiscreteIntRange private constructor(private val delegate: ContinuousRange<Int>) {
```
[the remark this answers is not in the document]
popematt: > Rename file to match class?

Good catch!

ion-schema/src/main/kotlin/com/amazon/ionschema/model/ContinuousRange.kt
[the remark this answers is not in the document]
rmarrowstone: Dang. Oh well.

PR #252, opened 2023-05-05
Adds the framework pieces for an Ion Schema reader that uses the ISL model

popematt:
**Issue #, if available:**

**Description of changes:**

Adds interfaces and common utilities for implementing an `IonSchemaReader` that will read Ion into the new ISL model. One improvement in particular over the existing reader logic *aside from the fact that this operates on the new model) is that this is designed to report multiple errors in a single schema document instead of just throwing an exception at the first error.

None


desaikd (approved):
I really like the idea of having the readers like `TypeReader`, `ConstraintReader` 🚀

PR #254, opened 2023-05-09
Adds CLI tool for fixing transitive imports

### LINKED ISSUE #178: Stop imports of schemas from including transitive dependencies

popematt:
**Issue #, if available:**

Fixes #178 

**Description of changes:**

Adds a tool to the `ion-schema-kotlin` CLI for fixing transitive imports.

To use it, you just have to check out `ion-schema-kotlin` and run `./ion-schema-cli repair fix-transitive-imports ...` and include the directory where your schemas are stored.

This uses string manipulation (rather than DOM manipulation) to actually apply the changes so that comments and whitespace in the original file don't get clobbered.

This PR looks really big, but 28 of the files are test fixtures. There's a set of schema files that need to be fixed (test input), and a corrected version using each of the three import reconciliation strategies (expected outputs).

None

### DIFF
--- a/cli/src/test/kotlin/com/amazon/ionschema/cli/commands/repair/FixTransitiveImportsCommandTest.kt
+++ b/cli/src/test/kotlin/com/amazon/ionschema/cli/commands/repair/FixTransitiveImportsCommandTest.kt
@@ -1,0 +1,106 @@
+package com.amazon.ionschema.cli.commands.repair
+
+import com.amazon.ion.system.IonSystemBuilder
+import com.amazon.ionschema.cli.IonSchemaCli
+import com.amazon.ionschema.cli.commands.repair.TransitiveImportRewriter.ImportStrategy
+import org.junit.jupiter.api.Assertions
+import org.junit.jupiter.api.BeforeAll
+import org.junit.jupiter.api.DynamicNode
+import org.junit.jupiter.api.DynamicTest.dynamicTest
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.TestFactory
+import org.junit.jupiter.api.TestInstance
+import java.io.File
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+import java.util.stream.Stream
+
+class FixTransitiveImportsCommandTest {
+
+    @Nested
+    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+    inner class NoWildcards {
+        @BeforeAll
+        fun runCommand() = runCommand(ImportStrategy.NoWildcards)
+        @TestFactory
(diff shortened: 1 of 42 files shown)


cli/src/test/kotlin/com/amazon/ionschema/cli/commands/repair/FixTransitiveImportsCommandTest.kt
```diff
@@ -0,0 +1,106 @@
...
+                    //     cli/src/test/resources/FixTransitiveImportsCommand/input/ \
+                    //     -d cli/src/test/resources/FixTransitiveImportsCommand/output-<STRATEGY>/
+                    testNodes.add(
+                        dynamicTest("Check $relativeFile are text equivalent", actualFile.toURI()) {
+                            Assertions.assertEquals(
+                                expectedFile.readText(),
+                                actualFile.readText(),
+                            )
+                        }
+                    )
```
tgregg, line 101: How are failures here presented to the user? I think it would be great if they failed separately from the tests that are expected not to be brittle. Can we group them together into something named obviously, like "PossiblyBrittleTests"?

tgregg (approved):
Nice tool

cli/src/test/kotlin/com/amazon/ionschema/cli/commands/repair/FixTransitiveImportsCommandTest.kt
popematt: Failures here are only presented to someone developing the tool. Not to a user of the tool. The test cases are named, and the potentially brittle ones are called `"Check $relativeFile are text equivalent"`. There is no risk of them failing unless you are messing with the source code of the tool and changing the text output.

Maybe brittle isn't the right word for this.

tgregg: Yeah I understand that. Still, as a developer, I find it nice to have tests that may fail in the future due to external factors marked explicitly. Could they start failing if IonJava's text writer changed its whitespacing strategy, for example?

popematt: Added a "⚠ Fragile ⚠" prefix to the name of the test cases.

PR #268, opened 2023-05-26
Make schema_footer optional

popematt:
**Issue #, if available:**

Followup for amazon-ion/ion-schema#118

**Description of changes:**

* Updates Ion Schema Kotlin so that the footer is optional and so that the footer is the last element of the schema and all trailing values (past the footer) are ignored.
* Updates the `ion-schema-tests` and `ion-schema-schemas` submodules

amazon-ion/ion-schema#118
 amazon-ion/ion-schema-tests#50
 amazon-ion/ion-schema-schemas#12

### DIFF
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/SchemaImpl_2_0.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/SchemaImpl_2_0.kt
@@ -47,7 +47,6 @@ internal class SchemaImpl_2_0 internal constructor(
     private val schemaSystem: IonSchemaSystemImpl,
     schemaContent: Iterable<IonValue>,
     override val schemaId: String?,
-
 ) : SchemaInternal {
 
     override val isl: IonDatagram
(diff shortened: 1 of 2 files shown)


ion-schema/src/main/kotlin/com/amazon/ionschema/internal/SchemaImpl_2_0.kt
```diff
@@ -92,6 +91,8 @@ internal class SchemaImpl_2_0 internal constructor(
         schemaContent.mapTo(isl) { it.clone() }
             .markReadOnly()
             .forEach {
+                // If we've already seen the footer, do nothing
+                if (foundFooter) return@forEach
```
tgregg, line 95: Should we look for another version marker and reenter the parsing loop if we find one? Or just stop processing?

tgregg (approved):

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/SchemaImpl_2_0.kt
popematt: >Should we look for another version marker and reenter the parsing loop if we find one?

No, I don't think so. There's a bunch of issues that make it undesirable to handle in the Ion Schema System right now. Basically, if we start parsing again, is the data part of the same schema, or is it a new schema? If it's part of the same schema, then we need to deal with the fact that we could have two different ISL versions contributing to the same schema (and this library is not in a state to handle that right now). If it's a different schema, then we have a situation where there are two schemas that came from one schema ID, and that's something that we don't want to allow.

I think the ideal long term solution would be to finish parsing one schema, and then return any remaining elements to the user. E.g., return a `class SchemaReadOutput(val schema: SchemaDocument, val unconsumedValues: Iterable<IonValue>)` or something like that—basically, we allow the user to decide whether to continue looking for more schemas in the Ion stream.

PR #273, opened 2023-05-31
Adds changes to read full schema document

popematt:
**Issue #, if available:**

#256 

**Description of changes:**

- Changes the signature of `readSchema` so that it returns a `Pair<SchemaDocument, Iterator<IonValue>>`—essentially, it returns the `SchemaDocument` and any unread `IonValue`s.
- Adds some utility functions for identifying top level content (header, footer, type, open content).
- Adds implementation of `readSchema` with a lot of tests.
- Fixes a bug with the `reportError` function that showed up while testing `readSchema`. Basically, fast-fail wasn't working properly because it was only throwing on fast fail without adding the error to the ReadContext.
- Another incidental change is adding a useful `toString()` for `Bag`. I realized that the default impl was giving me things like `Bag@398c0a`, so I changed it to something useful.

None.

### DIFF
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/reader/IonSchemaReader.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/reader/IonSchemaReader.kt
@@ -21,7 +21,7 @@ interface IonSchemaReader {
      * the given Ion. Reporting all errors is a best-effort attempt because the presence of one error can mask other
      * errors.
      */
-    fun readSchema(document: List<IonValue>, failFast: Boolean = false): IonSchemaResult<SchemaDocument, List<ReadError>>
+    fun readSchema(document: Iterable<IonValue>, failFast: Boolean = false): IonSchemaResult<Pair<SchemaDocument, Iterator<IonValue>>, List<ReadError>>
 
     /**
      * Reads a [SchemaDocument], throwing an [InvalidSchemaException][com.amazon.ionschema.InvalidSchemaException]
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/reader/IonSchemaReader.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/reader/IonSchemaReader.kt
@@ -21,7 +21,7 @@ interface IonSchemaReader {
      * the given Ion. Reporting all errors is a best-effort attempt because the presence of one error can mask other
      * errors.
      */
-    fun readSchema(document: Iterable<IonValue>, failFast: Boolean = false): IonSchemaResult<Pair<SchemaDocument, Iterator<IonValue>>, List<ReadError>>
+    fun readSchema(document: Iterable<IonValue>, failFast: Boolean = false): IonSchemaResult<SchemaDocument, List<ReadError>>
 
     /**
      * Reads a [SchemaDocument], throwing an [InvalidSchemaException][com.amazon.ionschema.InvalidSchemaException]
--- a/ion-schema/src/test/kotlin/com/amazon/ionschema/reader/IonSchemaReaderV2_0Tests.kt
+++ b/ion-schema/src/test/kotlin/com/amazon/ionschema/reader/IonSchemaReaderV2_0Tests.kt
@@ -1,0 +1,324 @@
+package com.amazon.ionschema.reader
+
+import com.amazon.ion.system.IonSystemBuilder
+import com.amazon.ionschema.InvalidSchemaException
+import com.amazon.ionschema.IonSchemaVersion
+import com.amazon.ionschema.model.Constraint
+import com.amazon.ionschema.model.DiscreteIntRange
+import com.amazon.ionschema.model.ExperimentalIonSchemaModel
+import com.amazon.ionschema.model.HeaderImport
+import com.amazon.ionschema.model.NamedTypeDefinition
+import com.amazon.ionschema.model.SchemaDocument
+import com.amazon.ionschema.model.TypeArgument
+import com.amazon.ionschema.model.TypeDefinition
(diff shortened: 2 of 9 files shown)


ion-schema/src/main/kotlin/com/amazon/ionschema/reader/IonSchemaReader.kt
```diff
@@ -21,7 +21,7 @@ interface IonSchemaReader {
      * the given Ion. Reporting all errors is a best-effort attempt because the presence of one error can mask other
      * errors.
      */
-    fun readSchema(document: List<IonValue>, failFast: Boolean = false): IonSchemaResult<SchemaDocument, List<ReadError>>
+    fun readSchema(document: Iterable<IonValue>, failFast: Boolean = false): IonSchemaResult<Pair<SchemaDocument, Iterator<IonValue>>, List<ReadError>>
```
desaikd, line 24: (Suggestion) Add a comment for what the `Pair<SchemaDocument, Iterator<IonValue>>` indicates here

ion-schema/src/test/kotlin/com/amazon/ionschema/reader/IonSchemaReaderV2_0Tests.kt
```diff
@@ -0,0 +1,324 @@
...
+        )
+        val iterator = result.unwrap().second
+        assertEquals(ION.singleValue("abc"), iterator.next())
+        assertEquals(ION.singleValue("def"), iterator.next())
+        assertEquals(ION.singleValue("\$ion_schema_1_0"), iterator.next())
+        assertFalse(iterator.hasNext())
+    }
+
+    @Test
+    fun `readSchema can read a schema with open content`() {
```
desaikd, line 166: Should we add an example for top level open content between schema header and version marker?

ion-schema/src/test/kotlin/com/amazon/ionschema/reader/IonSchemaReaderV2_0Tests.kt
```diff
@@ -0,0 +1,324 @@
...
+                    ),
+                )
+            ),
+            result.unwrap().first
+        )
+        assertFalse(result.unwrap().second.hasNext())
+    }
+
+    @Test
+    fun `readSchema can read a schema with a footer and returns an iterator with all of the content after the footer`() {
```
desaikd, line 140: (Question) Should we be returning the iterator for all the content thats not part of schema? `readSchema` to me doesn't seem like ti should return anything thats not part of the schema. 
I feel we could provide an alternative method maybe `readSchemaDocument`/`iterateSchemaDocument` which can return these non schema contents.

ion-schema/src/test/kotlin/com/amazon/ionschema/reader/IonSchemaReaderV2_0Tests.kt
```diff
@@ -0,0 +1,324 @@
...
+        assertEquals(ION.singleValue("abc"), iterator.next())
+        assertEquals(ION.singleValue("def"), iterator.next())
+        assertEquals(ION.singleValue("\$ion_schema_1_0"), iterator.next())
+        assertFalse(iterator.hasNext())
+    }
+
+    @Test
+    fun `readSchema can read a schema with open content`() {
+        val ionStream = ION.loader.load(
+            """
```
desaikd, line 168: (Question) If we decide to return the contents that are not part of schema then should we return something that is before the version marker?
i.e. from the given test should we also return `foo` in the iterator?

ion-schema/src/test/kotlin/com/amazon/ionschema/reader/IonSchemaReaderV2_0Tests.kt
```diff
@@ -0,0 +1,324 @@
...
+            foo
+            ${'$'}ion_schema_2_0
+            bar // Only bar is open content. Both 'foo' and 'baz' are outside the schema. 
+            schema_footer::{}
+            baz
+        """
+        )
```
desaikd, line 175: (Question) If we decide to return the contents that are not part of schema then should we return something that is before the version marker?
i.e. from the given test should we also return foo in the iterator?

ion-schema/src/test/kotlin/com/amazon/ionschema/reader/IonSchemaReaderV2_0Tests.kt
popematt: I don't think it matters, but I'll add one.

ion-schema/src/test/kotlin/com/amazon/ionschema/reader/IonSchemaReaderV2_0Tests.kt
popematt: I don't have a good answer, so I'm going to punt on this for now and change it back so that `readSchema` will only return the SchemaDocument.

ion-schema/src/main/kotlin/com/amazon/ionschema/reader/IonSchemaReader.kt
popematt: No longer applicable because I'm rolling back the change to the return type.

ion-schema/src/test/kotlin/com/amazon/ionschema/reader/IonSchemaReaderV2_0Tests.kt
desaikd: I think its nice to include all top level open content positions but I guess `ion-schema-tests` might already have those tests.

desaikd (approved):

PR #274, opened 2023-06-02
Ease-of-use improvements for ISL model

### LINKED ISSUE #264: Ion Schema model uses lists where it should use a set
The order of the arguments for `one_of`, `all_of`, `any_of`, `valid_values`, `timestamp_offset`, and `contains` does not change the behavior of the schema. These constraints should use a `Set` in the Ion Schema model so that `equals()` properly reflects their functional equivalence.

popematt:
**Issue #, if available:**

Fixes #264 

**Description of changes:**

By Commit:

* [Use sets instead of lists in the ISL model](https://github.com/amazon-ion/ion-schema-kotlin/commit/364933ce5147ed58061e50fa2f5dd32d51c1bc13)
  * Renames `TypeArgumentList` to `TypeArguments`
  * Adds `mapToSet()` convenience function (since the kotlin standard library `map` function maps to a list).
  * Changes logic constraints, contains, valid values, timestamp offset constraints to use a Set
  * Changes imports in schema header to be a set.
* [Makes ContinuousRange.Limit.Unbounded an object instead of a class](https://github.com/amazon-ion/ion-schema-kotlin/commit/e3203f29eaacf8bf0f6d3092fb2fb869be0d8eb7)
  * We only ever need one instance of `Unbounded`, so it makes sense for it to be an `object` instead of a `class`. 
* [Refactor SchemaDocument.Item](https://github.com/amazon-ion/ion-schema-kotlin/commit/b07860687fad338819e4ae5c1452d0821c982ec7)
  * Rename `Item` to `Content`, change to interface, and un-nest all members.
  * Get rid of `SchemaDocument.Item.Type` in favor of having `NamedTypeDefinition` implement `Content` directly.
  * `SchemaDocument.Item.Header` is now moved/renamed to `SchemaHeader`, and similarly with the footer.

None

### DIFF
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/model/SchemaDocument.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/model/SchemaDocument.kt
@@ -32,7 +32,7 @@ data class SchemaDocument(
         data class Type(val value: NamedTypeDefinition) : Item()
 
         data class Header(
-            val imports: List<HeaderImport> = emptyList(),
+            val imports: Set<HeaderImport> = emptySet(),
             val userReservedFields: UserReservedFields = UserReservedFields(),
             val openContent: OpenContentFields = emptyBag()
         ) : Item()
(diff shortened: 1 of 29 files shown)


ion-schema/src/main/kotlin/com/amazon/ionschema/model/SchemaDocument.kt
```diff
@@ -26,19 +25,14 @@ data class SchemaDocument(
     }
 
     /**
-     * Represents a top-level item in a schema document.
+     * Represents a top-level values in a schema document.
```
tgregg, line 28:
```suggestion
     * Represents a top-level value in a schema document.
```

tgregg (approved):

PR #282, opened 2023-08-02
Adds utility to get all symbol text declared in one or more schemas

popematt:
**Issue #, if available:**

None.

**Description of changes:**

Adds a utility that will get all of the symbols defined for types in schemas (i.e. all of the expected/possible field names, annotations, and valid values that the types define).

None.

### DIFF
--- a/ion-schema/src/test/kotlin/com/amazon/ionschema/util/SchemaSymbolsUtilTest.kt
+++ b/ion-schema/src/test/kotlin/com/amazon/ionschema/util/SchemaSymbolsUtilTest.kt
@@ -1,0 +1,147 @@
+package com.amazon.ionschema.util
+
+import com.amazon.ionschema.IonSchemaSystemBuilder
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+
+class SchemaSymbolsUtilTest {
+
+    companion object {
+        @JvmField
+        val ION_SCHEMA_2_0_SYMBOLS = setOf(
+            // Version marker
+            "\$ion_schema_2_0",
+            // Built in types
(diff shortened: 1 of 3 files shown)


ion-schema/src/test/kotlin/com/amazon/ionschema/util/SchemaSymbolsUtilTest.kt
```diff
@@ -0,0 +1,147 @@
...
+        assertEquals(ION_SCHEMA_2_0_SYMBOLS.sorted(), symbols.sorted())
+    }
+
+    @Test
+    fun `getSymbolsTextForSchema should get all expected symbols in a schema`() {
+        val schema = """
+            ${'$'}ion_schema_2_0
+            type::{
+              name: foo_type,
+              annotations: closed::[foo],
```
desaikd, line 108: Should we add tests for `annotations` that is defined using [standard syntax](https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec#standard-syntax)? 
```
annotations: { element: { valid_values: [foo, bar] } }  
```
Also, we might want to explicitly call out in doc comments if we are supporting `annotations` with standard syntax or simple syntax or both.

desaikd (approved):

ion-schema/src/test/kotlin/com/amazon/ionschema/util/SchemaSymbolsUtilTest.kt
popematt: It doesn't really matter, actually. The syntax is different, but the model representation is the same, and the implementation inspects the ISL model rather than the raw ISL.

PR #285, opened 2023-09-21
Add optimization to fail-fast on annotations

popematt:
**Issue #, if available:**

None

**Description of changes:**

Allows users to configure their schema system to fail early if a value does not match the annotations constraint for a type.
This is useful for cases where people use annotations to indicate the type of a value. For example:
```
type::{
  name: Address,
  one_of: [
    {
      annotations: required::closed::[PoBox],
      fields: {
        addressee: string,
        box_no: int,
        zip: int,
      }
    },
    {
      annotations: required::closed::[StreetAddress],
      fields: {
        addressee: string,
        street: string,
        city: string,
        state: string,
        zip: int,
      }
  ],
}
```
In this case, if a struct is annotated with `PoBox`, then there's not much point in checking it against the fields of the `StreetAddress` type. This has the benefit of reducing noise in the violations messages (why bother telling me how my `PoBox` value fails to be a `StreetAddress`—I already know it's not a `StreetAddress`), and improving performance (when calling `Type.validate()`) because it short circuits _some_ of the validation.

The exact performance improvement depends on the usage patterns, but in general, the more nesting there is inside a type/value that has an annotation, the larger the performance improvement will be.

None

### DIFF
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/TypeImpl.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/TypeImpl.kt
@@ -45,12 +45,12 @@ internal class TypeImpl(
         /**
          * Order in which constraints should be evaluated. Lower is first.
          *
-         * The [validate] function in this class has
+         * The [validate] function in this class has optional short-circuit logic for the `annotations` constraint, so
+         * `annotations` gets special treatment and will be evaluated first.
          */
         private val CONSTRAINT_EVALUATION_ORDER = mapOf(
             "annotations" to -1,
             // By default, all constraints are priority 0
-            "ordered_elements" to 1,
         )
         private val CONSTRAINT_PRIORITY_COMPARATOR = Comparator<Constraint> {
             a, b ->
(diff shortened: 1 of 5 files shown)


ion-schema/src/main/kotlin/com/amazon/ionschema/internal/TypeImpl.kt
```diff
@@ -40,6 +41,22 @@ internal class TypeImpl(
     private companion object {
         private val ION = IonSystemBuilder.standard().build()
         private val ANY = ION.newSymbol("any")
+
+        /**
+         * Order in which constraints should be evaluated. Lower is first.
+         *
+         * The [validate] function in this class has
```
desaikd, line 48: This seems incomplete.

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/TypeImpl.kt
```diff
@@ -40,6 +41,22 @@ internal class TypeImpl(
...
+         */
+        private val CONSTRAINT_EVALUATION_ORDER = mapOf(
+            "annotations" to -1,
+            // By default, all constraints are priority 0
+            "ordered_elements" to 1,
```
desaikd, line 53: This could use some comment on why `ordered_elements` is set to 1 and `annotations` to -1.

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/TypeImpl.kt
popematt: Fixed.

ion-schema/src/main/kotlin/com/amazon/ionschema/internal/TypeImpl.kt
popematt: Removed `ordered_elements`. Added a comment about why `annotations` gets special treatment.

desaikd (approved):

PR #286, opened 2023-09-22
Improves validation performance of types defined in the spec

popematt:
**Issue #, if available:**

None

**Description of changes:**

I discovered recently that the types defined in the spec are (subjectively) slow-ish for validation. I tried an alternative implementation, and ended with a pretty good improvement.

This change:
* Introduces `SpecType` (and `SpecTypes`, an object for holding all `SpecType` instances)
* Removes, `SchemaCore`, `TypeBuiltIn`, `TypeBuiltInImpl`, `TypeIon`, and `TypeCore`.
* Includes minor refactoring of some signatures to reflect that `SpecType` is effectively replacing `TypeBuiltIn`.

I could be convinced to rename `SpecType` to `TypeBuiltIn` so that the diff is smaller.

My informal performance testing revealed some really big improvements.  I chose some representative spec types that corresponded to 0, 1, "many", and "all" Ion types, and I created test cases that used the spec types to validate some sample data (one value of each Ion type except for clob and sexp). These results were measured by running warmups until the time was stable, using an "[autorange](https://docs.python.org/3/library/timeit.html#timeit.Timer.autorange)" approach to determine the number of invocations in each sample, and then collecting and averaging the ops/s for 30 samples.

|  Type   | Before (ops/s) | After (ops/s) | Difference |
|:-------:|---------------:|--------------:|-------:|
| nothing |         134126 |      14928349 | +11030% |
|   any   |         153279 |      44165490 | +28713% |
|  $any   |         135234 |      57456489 | +42386% |
| number  |         657438 |      11096099 |  +1587% |
| $number |         620417 |      12594188 |  +1930% |
|   int   |        2043462 |       6880997 |   +236% |
|  $int   |        1957613 |       9165562 |   +368% |

 I tested with both ISL 1.0 and 2.0, and the results were basically the same, but this is especially impactful for ISL 1.0, because any user type definition that does not have the `type` constraint has an implied `type: any`.

None.


ion-schema/src/test/kotlin/com/amazon/ionschema/TypeTest.kt
```diff
@@ -64,7 +64,6 @@ class TypeTest {
         val iter = violations.iterator()
         val violation = iter.next()
         assertEquals("type_mismatch", violation.code)
-        assertEquals("type", violation.constraint?.fieldName)
```
desaikd, line 67: (Question) Why was this removed? I couldn't find any changes in `Violation` structure for this PR.

popematt: The old way of creating the spec-defined types involved reading an Ion struct and then extracting the symbol from the struct. This meant that the symbols had the field name "type". See https://github.com/amazon-ion/ion-schema-kotlin/blob/a3fabe9a472095b23a2aa16c810da3f5b26734c4/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/SchemaCore.kt#L86-L118

Now we create the symbol directly instead, which means that the symbol no longer has a field name. (This whole thing is a weird consequence of the unfortunate decision that an `IonValue` knows its own field name.) Anyway, this lack of field name causes no difference in the violation messages, and the violation constraint field name is not actually used anywhere in the library.

desaikd (approved):

PR #289, opened 2023-09-27
Adds skeleton for ISL Writer

popematt:
**Issue #, if available:**

None

**Description of changes:**

This is the start of a writer for the Ion Schema model. It includes interfaces for a schema writer, type writer, constraint writers, and implementations of `HeaderWriter` and `FooterWriter`.

There's also a little bugfix in `IonSchemaResult`.

None

### DIFF
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/writer/internal/TypeWriter.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/writer/internal/TypeWriter.kt
@@ -1,0 +1,38 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.amazon.ionschema.writer.internal
+
+import com.amazon.ion.IonWriter
+import com.amazon.ionschema.model.DiscreteIntRange
+import com.amazon.ionschema.model.ExperimentalIonSchemaModel
+import com.amazon.ionschema.model.NamedTypeDefinition
+import com.amazon.ionschema.model.TypeArgument
+import com.amazon.ionschema.model.TypeArguments
+import com.amazon.ionschema.model.VariablyOccurringTypeArgument
+
+@ExperimentalIonSchemaModel
+internal interface TypeWriter {
+
+    /**
+     * Writes a [NamedTypeDefinition] to the given [IonWriter].
+     */
+    fun writeNamedTypeDefinition(ionWriter: IonWriter, namedTypeDefinition: NamedTypeDefinition)
+
+    /**
+     * Writes a [TypeArgument] to the given [IonWriter].
+     */
+    fun writeTypeArg(ionWriter: IonWriter, typeArg: TypeArgument)
+
+    /**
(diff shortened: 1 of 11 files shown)


ion-schema/src/main/kotlin/com/amazon/ionschema/writer/internal/TypeWriter.kt
```diff
@@ -0,0 +1,38 @@
...
+     */
+    fun writeTypeArg(ionWriter: IonWriter, typeArg: TypeArgument)
+
+    /**
+     * Writes a [VariablyOccurringTypeArgument] to the given [IonWriter].
+     */
+    fun writeVariablyOccurringTypeArg(ionWriter: IonWriter, varTypeArg: VariablyOccurringTypeArgument, elideOccursValue: DiscreteIntRange)
+
+    /**
+     * Writes a [TypeArguments] to the given [IonWriter].
```
desaikd, line 33:
```suggestion
     * Writes a [TypeArgument]s to the given [IonWriter].
```

ion-schema/src/main/kotlin/com/amazon/ionschema/writer/internal/TypeWriter.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.amazon.ion.IonWriter
+import com.amazon.ionschema.model.DiscreteIntRange
+import com.amazon.ionschema.model.ExperimentalIonSchemaModel
+import com.amazon.ionschema.model.NamedTypeDefinition
+import com.amazon.ionschema.model.TypeArgument
+import com.amazon.ionschema.model.TypeArguments
+import com.amazon.ionschema.model.VariablyOccurringTypeArgument
+
+@ExperimentalIonSchemaModel
+internal interface TypeWriter {
```
desaikd, line 15: (Question) This interface contains both type argument and type definition write methods. I am not sure how these methods will be used though. If you can provide an example that would be helpful.

ion-schema/src/main/kotlin/com/amazon/ionschema/writer/internal/TypeWriter.kt
popematt: There's a type alias called `TypeArguments` so this is correct, despite how it looks.

ion-schema/src/main/kotlin/com/amazon/ionschema/writer/internal/TypeWriter.kt
popematt: The schema writer will call the type definition method to write top-level types in a schema document. Constraint writers can call the type argument method if they have a type argument.

They could be in separate interfaces, because they are called from mutually exclusive locations, but since it's an internal interface, I don't think it's a big deal whether they're together or separate.

desaikd: Sure. It wouldn't matter for an internal interface but since this PR doesn't add any usage of this interface, I was just curious how this is intended to be used. 
IMO, it might be worth adding interface definition with at least one place where its implemented or an example/summary of how the interface definition will be used in the PR description.

desaikd (approved):

PR #295, opened 2023-10-26
Update reader APIs

popematt:
**Issue #, if available:**

None

**Description of changes:**

Updates the reader APIs so that the user doesn't have to know the schema version ahead of time. Now the specific version is an implementation detail, and the user interacts with the reader by calling `IonSchemaReader.readSchema()`.

None

### DIFF
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/reader/internal/IonSchemaReaderV2_0.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/reader/internal/IonSchemaReaderV2_0.kt
@@ -1,4 +1,7 @@
-package com.amazon.ionschema.reader
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.amazon.ionschema.reader.internal
 
 import com.amazon.ion.IonValue
 import com.amazon.ionschema.InvalidSchemaException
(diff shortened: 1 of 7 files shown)


ion-schema/src/main/kotlin/com/amazon/ionschema/reader/internal/IonSchemaReaderV2_0.kt
```diff
@@ -9,20 +12,10 @@ import com.amazon.ionschema.model.ExperimentalIonSchemaModel
...
-import com.amazon.ionschema.reader.internal.isFooter
-import com.amazon.ionschema.reader.internal.isHeader
-import com.amazon.ionschema.reader.internal.isTopLevelOpenContent
-import com.amazon.ionschema.reader.internal.isType
-import com.amazon.ionschema.reader.internal.readCatching
 import com.amazon.ionschema.util.IonSchemaResult
 
 @OptIn(ExperimentalIonSchemaModel::class)
-class IonSchemaReaderV2_0 : IonSchemaReader {
+internal object IonSchemaReaderV2_0 : VersionedIonSchemaReader {
```
desaikd, line 18: (Question) Why did we move this to be internal? Which reader is available to access by user?

popematt: The user doesn't directly access any reader. The actual reader is stateless, so there's no advantage to holding a reference to one. Instead, the user calls `IonSchemaReader.readSchema()` or  `IonSchemaReader.readSchemaOrThrow()` (both of which are static functions). This hides the implementation details a little better and will (eventually) be able to read both 1.0 and 2.0 without the user having to know the version ahead of time.

desaikd (approved):

PR #296, opened 2023-11-13
Remove extraneous clones in TypeNamed and ViolationChild

popematt:
*Issue #, if available:*

None.

*Description of changes:*

I changed the implementation of `ViolationChild` so that it's not unnecessarily creating clones of `IonValue`s.
Also changed the implementation of `TypeNamed` so that it doesn't create a new struct _and_ clone the entire type definition every time it validates a value.

My informal performance testing shows a significant improvement. I created a representative schema:
```
type::{ name: A, type: int }
type::{ name: B,  type: string }
type::{ name: C, type: $any }
type::{
  name: abc,
  fields: {
    a: A,
    b: B,
    c: C,
  }
}
```
And tested it using a set of values that were all valid for the type `abc` and another set that were all invalid. These results were measured by running warmups until the time was stable, using an "[autorange](https://docs.python.org/3/library/timeit.html#timeit.Timer.autorange)" approach to determine the number of invocations in each sample, and then collecting and averaging the ops/s for 30 samples.

| Dataset | Before (ops/s) | After (ops/s) | Change |
|:-------:|---------------:|--------------:|-------:|
|  valid  |        1375113 |       2097905 |    52% |
| invalid |         865061 |       1065709 |    23% |

### DIFF
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/Violations.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/Violations.kt
@@ -15,7 +15,6 @@
 
 package com.amazon.ionschema
 
-import com.amazon.ion.IonStruct
 import com.amazon.ion.IonValue
 import com.amazon.ionschema.internal.util.truncate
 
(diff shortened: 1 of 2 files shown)


zslayton (approved):

tgregg (approved):

desaikd (approved):
🚀

Issue #307, opened 2025-09-11
$null_or is hard to use correctly and the errors are misleading


mtcalvin:
Consider this schema:

```
$ion_schema_2_0
type::{
  name: foo,
  type: $null_or::string,
  codepoint_length: range::[1, 10],
}
```

The type is `$null_or::string`, however, `null` can never be valid because of the `codepoint_length` constraint.

ISL Kotlin and Rust print different errors in this case when presented with `null`:

Rust:
```
type_mismatched | expected [String, Symbol] but found null
```

Kotlin:
```
Validation failed:
- not applicable for type null
```

Neither one of these intuitively explains why `null` isn't valid; one must consult the ISL specification to deduce that the problem is that `codepoint_length` cannot be applied to `null` because `null` has no length by definition.  The fix here is to do something like this:

```
$ion_schema_2_0
type::{
  name: foo,
  type: $null_or::{
    type:string,
    codepoint_length: range::[1, 10]
  }
}
```

However this is not obvious to most casual users of ISL.  Most violations of this type (null cannot have length constraint applied) are highly likely to be caused by this misunderstanding when defining the schema rather than the data itself being invalid (the developer likely intended for `null` to be valid or else they wouldn't have added the `$null_or` in the first place).

I think the error messaging here could be improved to more accurately direct users to the source of the problem -- for example, if a `type:$null_or::something` appears at the same level as any other constraint, it's very likely a mistake.  Something like `Value "null" is not valid because the type defines codepoint_length constraints which are only valid for non-null values` would be much more clear to understand.

austnwil:
Error messages on null violations were improved in `ion-schema-kotlin`. See the linter issue linked above to track desire for a more comprehensive tool that can catch mistakes/pitfalls like this.

PR #308, opened 2025-09-23
Report descriptive error when a type definition with type: $null_or:: cannot accept null due to a conflicting constraint

austnwil:
Addresses #307 

## Description of changes:

This change creates more descriptive error messages in cases where a schema uses a `$null_or::` annotation on the `type:` field of a type definition that also includes other constraints for which null are never valid. This can address some confusion where a user should have instead used `$null_or` on the type definition itself and point them in the right direction.

Previously, a schema like the following:

```
$ion_schema_2_0
type::{
    name: mystring,
    type: $null_or::string,
    codepoint_length: range::[ 1, 10 ]
}
```

would report the following error when validating null against `mystring`:

```
Validation failed:
- not applicable for type null
```

because null has no codepoint_length and therefore is invalid for this constraint. However, this error is vague and unhelpful for someone figuring out why this doesn't work.

The same schema will now report the following when validating null:

```
Validation failed:
- type cannot accept null. note: type attempts to accept null via $null_or but defines one or more constraints for which null are never valid - did you mean to use $null_or on the type definition itself?
  - constraint "codepoint_length" is not applicable for null values
```

Multiple violations due to constraints that do not accept null are grouped together. As an example, this java code:

```java
final var iss = IonSchemaSystemBuilder.standard().build();
final var is = IonSystemBuilder.standard().build();

final Schema schema = iss.newSchema("""
    $ion_schema_2_0
    type::{
        name: mystring,
        type: $null_or::string,
        annotations: closed::[ abc ],
        utf8_byte_length: range::[ min, 4096 ],
        regex: i::"^[a-z0-9_]+$",
    }
    """);

Type type = schema.getType("mystring");
final var ionValue = is.singleValue("abc::null");

Violations violations = type.validate(ionValue);
if (!violations.isValid()) {
    System.out.println(violations);
}
```

will output these validation errors:

```
Validation failed:
- type cannot accept null. note: type attempts to accept null via $null_or but defines one or more constraints for which null are never valid - did you mean to use $null_or on the type definition itself?
  - constraint "utf8_byte_length" is not applicable for null values
  - constraint "regex" is not applicable for null values
```

Other violations are still reported separately:

```
$ion_schema_2_0
type::{
    name: mystring,
    type: $null_or::string,
    codepoint_length: range::[ 1, 10 ],
    annotations: closed::[]
}
```

```
// validate not_allowed::null
Validation failed:
- found one or more unexpected annotations
- type cannot accept null. note: type attempts to accept null via $null_or but defines one or more constraints for which null are never valid - did you mean to use $null_or on the type definition itself?
  - constraint "codepoint_length" is not applicable for null values
```

### DIFF
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/CommonViolations.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/CommonViolations.kt
@@ -26,12 +26,12 @@ internal object CommonViolations {
     fun INVALID_TYPE(constraint: IonValue, value: IonValue) = Violation(
         constraint,
         "invalid_type",
-        "not applicable for type %s".format(value.type.toString().toLowerCase())
+        "values of type %s are never valid for types defining %s constraints".format(value.type.toString().toLowerCase(), constraint.fieldName)
     )
 
     fun NULL_VALUE(constraint: IonValue) = Violation(
         constraint,
         "null_value",
-        "not applicable for null values"
+        "null values are never valid for types defining %s constraints".format(constraint.fieldName)
     )
 }
--- a/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/TypeImpl.kt
+++ b/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/TypeImpl.kt
@@ -108,23 +108,20 @@ internal class TypeImpl(
         val typeWantsToAcceptNull = isl.get("type")?.hasTypeAnnotation("\$null_or") == true
         val incompatibleConstraintsWithNullOrIssues = Violation(
             isl,
-            "null_value",
-            "type cannot accept null. note: type attempts to accept null via \$null_or but defines one or more constraints which are only valid for non-null values - did you mean to use \$null_or on the type definition itself?"
+            "constraints_incompatible",
+            "type cannot accept null. note: type attempts to accept null via \$null_or but defines one or more constraints for which null are never valid - did you mean to use \$null_or on the type definition itself?"
         )
 
         fun validateConstraintAndSeparateNullViolations(constraint: Constraint) {
             val constraintIssues = Violations()
-            val constraintNullValueIssues = Violation(message = "constraint \"${constraint.name}\" is not compatible with type: \$null_or")
+            val constraintNotApplicableForNullViolation = Violation(message = "constraint \"${constraint.name}\" is not applicable for null values")
             constraint.validate(value, constraintIssues)
             constraintIssues.forEach {
                 if (it.code == "null_value") {
-                    constraintNullValueIssues.add(it)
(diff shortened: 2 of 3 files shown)


popematt (commented):
While your overall approach appears to work, I think it is more complex than it needs to be. We don't have to improve the messages _only_ when there is a `$null_or` annotation—we can improve the message for all null-related violations if it keeps things simpler.

Can you see if we can solve things by updating this function?
https://github.com/amazon-ion/ion-schema-kotlin/blob/7a9ee6d06e9cfdaa530310f5d9be9f5a52680d3b/ion-schema/src/main/kotlin/com/amazon/ionschema/internal/CommonViolations.kt#L32-L36

Also, are there any tests that need to be updated as a result of this change?

austnwil:
> We don't have to improve the messages _only_ when there is a `$null_or` annotation—we can improve the message for all null-related violations if it keeps things simpler.

I can definitely simplify this by updating the error message in CommonViolations.kt instead of the custom logic to detect if the user is using `$null_or` incorrectly in particular, but the error might be less helpful.

> Also, are there any tests that need to be updated as a result of this change?

This change doesn't affect whether or not any value is valid for a particular type, so there's nothing to really test.

I wanted to add tests, but the test suite doesn't seem capable of testing the actual violations emitted. It only allows testing that schemas and type definitions are correctly accepted or rejected as valid/invalid and that a particular type accepts or rejects a particular value in general.

austnwil:
I simplified the code and made null value and invalid type errors more descriptive in all cases, not just when the user attempts to use `$null_or` with a constraint that cannot accept null.

Below are some example error messages when validating a few different nulls against this schema:

```
$ion_schema_2_0
type::{
    name: mytext,
    type: $null_or::text,
    regex: "[a-z]+"
}
```

```
Validating null.symbol
Validation failed:
- expected type text, found null symbol
- null values are never valid for types defining regex constraints


Validating null.string
Validation failed:
- expected type text, found null string
- null values are never valid for types defining regex constraints


Validating null.int
Validation failed:
- expected type text, found null int
- values of type int are never valid for types defining regex constraints


Validating null
Validation failed:
- values of type null are never valid for types defining regex constraints


Validating null.null
Validation failed:
- values of type null are never valid for types defining regex constraints
```

These errors I believe are slightly less helpful than the previous error because they don't explicitly indicate to the user *why* using `$null_or` in this manner is incorrect and that you should instead use it on the type definition itself. However, I think they are still good enough to point people in the right direction.

popematt:
> These errors I believe are slightly less helpful than the previous error because they don't explicitly indicate to the user why using $null_or in this manner is incorrect and that you should instead use it on the type definition itself. However, I think they are still good enough to point people in the right direction.

I agree—and the ideal long term solution is an Ion Schema linter that will point out things that you are doing that are probably wrong.

popematt (approved):

